PR #3, opened 2017-01-31
Add build-env:create-project command 

greg-1-anderson:
Ported from https://github.com/stevector/pantheon-create-project.

See #7

### DIFF
--- a/src/Commands/BuildToolsCommand.php
+++ b/src/Commands/BuildToolsCommand.php
@@ -308,8 +308,11 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
     protected function siteHasMultidevCapability($site)
     {
         // Can our new site create multidevs?
-        // TODO: For some reason, this field is not being populated in some instances.
-        return true; // $site->get('max_num_cdes') > 0;
+        $settings = $this->get('settings');
+        if (!$settings) {
+            return false;
+        }
+        return $settings->max_num_cdes > 0;
     }
 
     /**
--- a/src/Commands/BuildToolsCommand.php
+++ b/src/Commands/BuildToolsCommand.php
@@ -875,13 +875,6 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
         $metadata = $this->getBuildMetadata($repositoryDir);
         $this->recordBuildMetadata($metadata, $repositoryDir);
 
-        // Create a new branch and commit the results from anything that may
-        // have changed. We presume that the source repository is clean of
-        // any unwanted files prior to the build step (e.g. after a clean
-        // checkout in a CI environment.)
-        $this->passthru("git -C $repositoryDir checkout -B $branch");
-        $this->passthru("git -C $repositoryDir add --force -A .");
-
         // Remove any .git directories added by composer from the set of files
         // being committed. Ideally, there will be none. We cannot allow any to
         // remain, though, as git will interpret these as submodules, which
@@ -899,6 +892,13 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
             ->name('.git')
             ->getIterator()
         );
+
+        // Create a new branch and commit the results from anything that may
+        // have changed. We presume that the source repository is clean of
+        // any unwanted files prior to the build step (e.g. after a clean
+        // checkout in a CI environment.)
+        $this->passthru("git -C $repositoryDir checkout -B $branch");
+        $this->passthru("git -C $repositoryDir add --force -A .");
 
         // Now that everything is ready, commit the build artifacts.
         $this->passthru("git -C $repositoryDir commit -q -m 'Build assets for $env_label.'");
@@ -967,11 +967,11 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
         // Push our changes back to the dev environment, replacing whatever was there before.
         $this->passthru('git push --force -q pantheon master');
 
-        // Once the build environment is merged, we do not need it any more
-        $this->deleteEnv($env, true);
-
         // Wait for the dev environment to finish syncing after the merge.
         $this->waitForCodeSync($preCommitTime, $site, 'dev');
+
+        // Once the build environment is merged, we do not need it any more
+        $this->deleteEnv($env, true);
     }
 
     /**
--- a/src/Commands/BuildToolsCommand.php
+++ b/src/Commands/BuildToolsCommand.php
@@ -920,8 +920,7 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
         // sftp mode before this sync is completed, then the converge that
         // sftp mode kicks off will corrupt the environment.
         if ($environmentExists) {
-            $target_env = $site->getEnvironments()->get($multidev);
-            $this->waitForCodeSync($preCommitTime, $site, $target_env);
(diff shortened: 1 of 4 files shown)


stevector:
@greg-1-anderson, shall we bring over the circle.yml pieces so that this functionality itself can be tested? We'd need a new github user like the @stevectorbot account I used over in https://github.com/stevector/pantheon-create-project

greg-1-anderson:
Yes, I think it would be a good idea to have some circle tests that use create-project to clone pantheon-systems/example-drops-8-composer and set up new GitHub, Pantheon and Circle projects with the create-project command.

It would probably be too meta to submitting a PR to the new GitHub account and hang out until the test passed. However, I think we could test the other commands in this tool by directly creating a multidev account and walking through all of the steps that the example.circle.yml does, and confirm the results.

greg-1-anderson:
Note also, though, that this PR is still a little ways from "done". There are a couple entire methods that still need to be ported from `bash`, and there are some variables that are not calculated &/or passed all the way through to the methods that use them. None of this code has been executed yet.

stevector:
Sounds good, let know if you want to talk through any of the pieces.

stevector:
@greg-1-anderson, it worked for me too! 

https://github.com/stevector/pantheon-create-project/pull/8/files
https://circleci.com/gh/stevector/pantheon-create-project/40
https://circleci.com/gh/stevectorbot/stevectorbot-40/tree/master

src/Commands/BuildToolsCommand.php
```diff
@@ -39,6 +43,413 @@ public function __construct()
...
+        // Look up the site UUID for the Pantheon dashboard link
+        $site = $this->getSite($site_name);
+        $siteInfo = $site->serialize();
+        $site_uuid = $siteInfo['id'];
+
+        $this->log()->notice('Created a new Pantheon site with UUID {uuid}', ['uuid' => $site_uuid]);
+
+        // Create a new README file to point to this project's Circle tests and the dev site on Pantheon
+        $circleBadge = "[![CircleCI](https://circleci.com/gh/$source.svg?style=svg)](https://circleci.com/gh/{$target_project})";
+        $pantheonBadge = "[![Pantheon {$target}](https://img.shields.io/badge/pantheon-{$target}-yellow.svg)](https://admin.dashboard.pantheon.io/sites/{$site_uuid}#dev/code)";
```
stevector: This command use the regular dashboard instead of the admin dashboard.

src/Commands/BuildToolsCommand.php
```diff
@@ -39,6 +43,413 @@ public function __construct()
...
+        $siteBadge = "[![Dev Site {$target}](https://img.shields.io/badge/site-{$target}-blue.svg)](http://dev-{$target}.pantheonsite.io/)";
```
stevector: Two of the generated shields failed here: https://github.com/stevectorbot/stevectorbot-40

greg-1-anderson: I guess you cannot have dashes in the badge text. Will need to decide what to do about that.

src/Commands/BuildToolsCommand.php
greg-1-anderson: Yeah, you're right. Thanks.

src/Commands/BuildToolsCommand.php
greg-1-anderson: Converting dashes to underscores works--although underscores render as spaces, it's probably close enough for badge purposes.

stevector:
@greg-1-anderson, we'll need to decide how to handle sites that don't have access to multidev.

greg-1-anderson:
@stevector The original template site README advises to switch to `dev` for everything by editing your `circle.yml`. We could auto-degrade to that on site creation if multidev was not available.

Do you happen to know how to detect multi-dev capable sites via Terminus? We could check the service level via `terminus site:info`, and hardcode the non-multi-dev service levels if there is nothing more direct.

stevector:
I'd really rather not try to rely on service level since we have so many Sandbox testing sites that we've manually flipped to multidev. I don't see a Terminus command for this specifically but I'm sure one could be written.

greg-1-anderson:
Is there any sort of web service API or other way to remotely detect if multidev is available? As a fallback, we could try to create one, but that would be slow and inelegant.

greg-1-anderson:
Maybe the best thing to do, failing an infallible way to detect multidev capability, would be to have an option to select whether the user wishes to use multidev or not. We could perhaps use the service level to provide the default value if the user does not explicitly specify it.

src/Commands/BuildToolsCommand.php
```diff
@@ -39,6 +43,460 @@ public function __construct()
...
+        $target_org = $github_org;
+        if (empty($github_org)) {
+            $createRepoUrl = 'user/repos';
+            $userData = $this->curlGitHub('user', [], $github_token);
+            $target_org = $userData['login'];
+        }
+        $target_project = "$target_org/$target";
+        $remote_url = "git@github.com:${target_project}.git";
+
+        $branch = ''; // TODO: allow non-master here?
```
stevector: Yes, I was talking with an agency today about Lightning. It'd be nice to run this command against https://github.com/pantheon-systems/example-drops-8-composer/pull/33/files

src/Commands/BuildToolsCommand.php
```diff
@@ -39,6 +43,460 @@ public function __construct()
...
+        $version = ''; // TODO: allow version selection here? (e.g. '~2' or '^1.5')
```
stevector: We could just have `$source_project` be passed in fully formed rather than construct it from multiple variables

greg-1-anderson: Yeah, I was thinking about going that route too. Sounds like a good plan.

src/Commands/BuildToolsCommand.php
greg-1-anderson: I think that would be `pantheon-systems/example-drops-8-composer/dev-lightning-fist-2`.

greg-1-anderson:
@stevector I pushed a commit to allow `org/project:dev-branch`, but I haven't tested it yet.

src/Commands/BuildToolsCommand.php
```diff
@@ -173,6 +594,163 @@ public function createBuildEnv(
...
+    {
+        // TODO: Detect WordPress sites and use wp-cli to install.
+        list($site, $env) = $this->getSiteEnv($site_env_id);
+
+        $this->log()->notice('Install site on {site}', ['site' => $site_env_id]);
+
+        // Set the target environment to sftp mode prior to installation
+        $this->connectionSet($env, 'sftp');
+
+        $command_line = "drush site-install --yes";
```
stevector: I made this branch a month ago but forgot to open the PR apparently: https://github.com/pantheon-systems/example-drops-8-composer/pull/60

Different repos (WordPress, Acquia Lightning, etc) may have totally different install steps. I think we should let the given repo provide its own CMS installation instructions.

greg-1-anderson:
OMG passing tests! With Circle CI inception!

greg-1-anderson:
I'm going to merge this PR as an EXPERIMENTAL feature; backwards compatibility is not guaranteed until the command is completed, at which point I will tag version 1.3.0. This will make it easy for folks to try this command, while still giving us the latitude to make breaking changes to get the API / options right.

PR #15, opened 2017-04-10
Better Handling of Multidev Branch Naming to Handle Multiple Underscores

populist:
There is a problem with the existing logic in creating Multidev branch naming that in cases where there are multiple underscores they don't properly get filtered into Multidev acceptable naming. FOr example:

underscore_name_fix
update_local_dev_readme

The updated replacement logic will filter out cases like these.


carsonblack:
I'm not listed as a reviewer, but, approved!

greg-1-anderson:
LGTM

PR #35, opened 2017-05-09
WIP:  Git provider refactor

rvtraveller:
This is a WIP.  I still have some cleanup to do (namely, putting in the build metadata the providers used and removing some hard coded lists).  Hoping to start conversation about general direction.  The idea is to factor out the git and CI providers into classes so multiple can be provided with build tools and the user can specify what providers they want to use.  So if someone wants to use Bitbucket + Circle, they can.  If someone wants to use GitLab + GitLab CI, they can.


greg-1-anderson:
Thanks for jumping in here -- the only reason that the code isn't already factored like this is that implementation started during Terminus 1.0, where autoloading is not supported.

It is my general intention to allow what you are enabling here. However, what I thought I would do is use [robo tasks](http://robo.li/getting-started/#tasks) to implement the different sorts of repository and CI providers. This would also allow us to leverage [robo configuration](http://robo.li/getting-started/#configuration) to pass settings from a configuration file directly into the provider named by the configuration item.

Is this something you'd be interested in contributing to?

rvtraveller:
The robo tasks and configuration are interesting.  Seems like it would be cleaner than what I have.  Depending on the scope, it could also be a fairly significant rewrite over what is here now.  It almost seems worthwhile to move the entire build metadata system into a robo config file which is read in for each repository rather than the current JSON file.  Although I suppose the first steps would be more towards the robo tasks.

Happy to contribute toward this.

greg-1-anderson:
The purpose of the build metadata system is to provide information about the current multidev environment to the quicksilver hooks, so they know which github repository and branch to push code back to. I chose json over yml for this to make the metadata easier to parse in the quicksilver hook.

There would be a bit of refactoring / rewriting to do here, indeed, but in essence, a Robo task is nothing more than a class that implements TaskInterface, which has but a single `run()` method. So, all you really need to do to convert existing code to tasks is make classes that are configured via setter methods, and then do all of their work in the `run()` method.

greg-1-anderson:
I think I'm going to start working on this now. My plan:

- [ ] Refactor the single commandfile into one file per command
- [ ] Refactor build-env:create-project into Robo tasks that just use `addCode()` to run existing functions.
- [ ] Create Robo tasks for CI, and move existing CircleCI code there
- [ ] Extend to add one other CI service, maybe Travis or Jenkins

greg-1-anderson:
I realized that it would be best if `@hook interact` was aware of the specific repository and CI objects in order to, at a minimum, provide specific messages about the authentication token needed by these services. Today, this method includes the URL to each specific service's page to create an authentication token. It is also possible that some services might need more than a single authentication value from the user. Also, although `@hook validate` does not do service-specific validation, it might be good to allow this in the future.

The implication of this is that this enhancement will necessarily look more similar to what was contributed here, as I won't be able to rely solely on Robo task configuration to select the desired services.

greg-1-anderson:
The master branch now has a provider implementation similar to what is in this PR. Thank you for the suggestions.

PR #9, opened 2017-03-13
WIP - Trying wordpress as source

stevector:
@greg-1-anderson, I expect this will break, but I'm not exactly sure how. Can you give me write access to this repo so that the PR can build with write access?

Issue #5, opened 2017-03-02
Testing strategy for permutations of external git hosts / CI services / CMS versions


stevector:
Over in https://github.com/pantheon-systems/example-drops-8-composer/issues/61#issuecomment-283417505 @greg-1-anderson highlighted some of the challenges we will have accommodating different services. I agree that tracking all of these permutations will be very complex and difficult.

I think we will benefit from writing out baseline expectations for any given permutation. And I think Behat would be of help here.

Imagine a Behat `.feature` file meant to lay out the expectations for `build-env:create-project`. I think it would look something like this:

```
Feature: create-project
In order to create a new site based on an existing example
As a Pantheon user
I need a command that will automatically connect an external repository, a CI service, Pantheon

Scenario: Basic project
Given a local, authenticated terminus connection
And authentication information for an external git host
And authentication information for an external CI service
And a Pantheon machine token to be used in the to-be-created automation setup

When I run pass the name of an example project to the build-env:create-project command

Then I will have created a new repository on the external git host
And I will have created a new site on Pantheon
And the CMS will have been installed on the Dev environment of the Pantheon Site
And a CI build will have passed on the external CI service.
```

We could then run this .feature file in a Behat suite per permutation where each suite swapped out different PHP classes that define the appropriate specifics. For instance, we would end up with a Behat context per git host with each context defining the same steps like `Given authentication information for an external git host`. Of course we would then want this actual Terminus plugin to have a similar set of classes (all implementing the same interface) to define the actual interaction with GitHub/Bitbucket/GitLab.

And yes, running all of the Behat suites would be incredible time consuming. I think we would rely on PRs to this repo running a small subset of permutations and then running the full set on some spaced out schedule.

stevector:
I should also add that I would expect Behat tests here would be supplemental to unit tests that could run faster and on every PR this this repo.

greg-1-anderson:
Once I get the basic circle.yml working here to test a single permutation, I think converting it to behat tests as described above would be a good idea. Behat has the nice capability of being able to run through a scenario multiple times, using data from an inline table for each pass through. That could be useful here. I'm not sure if there would be a way to set up a "quick run" that only stepped through some of the available fixture data.

[...]

PR #39, opened 2017-05-23
WIP - Support for WordPress example repo

stevector:
Here is a work-in-progress PR to replace https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/9

This PR should not be merged until

- [ ] There's a mechanism in place for running CI against multiple values of `SOURCE_COMPOSER_PROJECT` (My thought is to use a matrix of variable in Travis). Related to https://github.com/pantheon-systems/terminus-build-tools-plugin/issues/5
- [ ] The substance of this PR itself is reviewed (I'm greatly encouraged by how small this PR is)
- [ ] The WordPress example repo is moved from `stevector/example-wordpress-composer` to `pantheon-systems/example-wordpress-composer`, [a task which has its own set of issues](https://github.com/stevector/example-wordpress-composer/issues?q=is%3Aissue+is%3Aopen+label%3ABlocks-move-to-pantheon-systems).

### DIFF
--- a/src/Commands/BuildToolsCommand.php
+++ b/src/Commands/BuildToolsCommand.php
@@ -659,7 +659,7 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
      */
     protected function autodetectUpstream($siteDir)
     {
-        if (file_exists($siteDir . 'wp-cli.yml')) {
+        if (file_exists($siteDir . '/wp-cli.yml')) {
             return 'WordPress';
         }
         else {
--- a/src/Commands/BuildToolsCommand.php
+++ b/src/Commands/BuildToolsCommand.php
@@ -379,6 +379,7 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
                 'account-pass' => $circle_env['ADMIN_PASSWORD'],
                 'site-mail' => $circle_env['ADMIN_EMAIL'],
                 'site-name' => $circle_env['TEST_SITE_NAME'],
+                'site-url' => "https://dev-{$site_name}.pantheonsite.io"
             ];
             $this->doInstallSite("{$site_name}.dev", $composer_json, $site_install_options);
 
@@ -394,11 +395,11 @@ class BuildToolsCommand extends TerminusCommand implements SiteAwareInterface
             $this->configureCircle($target_project, $circle_token, $circle_env);
         }
         catch (\Exception $e) {
-            $ch = $this->createGitHubDeleteChannel("repos/$target_project", $github_token);
-            $data = $this->execCurlRequest($ch, 'GitHub');
-            if (isset($site)) {
-                $site->delete();
-            }
+//            $ch = $this->createGitHubDeleteChannel("repos/$target_project", $github_token);
+//            $data = $this->execCurlRequest($ch, 'GitHub');
+//            if (isset($site)) {
+//                $site->delete();
+//            }
             throw $e;
         }
         $this->log()->notice('Your new site repository is {github}', ['github' => "https://github.com/$target_project"]);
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
We also need to make an `Empty Upstream` for WordPress (and eventually Drupal 7 as well).

src/Commands/BuildToolsCommand.php
```diff
@@ -658,9 +659,13 @@ protected function expandSourceAliases($source)
      */
     protected function autodetectUpstream($siteDir)
     {
-        return 'Empty Upstream';
-        // or 'Drupal 7' or 'WordPress'
-        // return 'Drupal 8';
+        if (file_exists($siteDir . '/wp-cli.yml')) {
+            return 'WordPress';
```
greg-1-anderson: I created a new empty upstream w/ the framework set to 'wordpress' named 'Empty WordPress'

src/Commands/BuildToolsCommand.php
```diff
@@ -658,9 +659,13 @@ protected function expandSourceAliases($source)
...
+        }
+        else {
+            // This upstream works for Drupal 8 and Drupal 7.
```
greg-1-anderson: Actually, 'Empty Upstream' only works for Drupal 8. I made another upstream named 'Empty 7' for Drupal 7. Maybe test for "$siteDir/web/misc/drupal.js" as an indicator that the site is Drupal 7.

src/Commands/BuildToolsCommand.php
```diff
@@ -658,9 +659,13 @@ protected function expandSourceAliases($source)
      */
     protected function autodetectUpstream($siteDir)
     {
-        return 'Empty Upstream';
-        // or 'Drupal 7' or 'WordPress'
-        // return 'Drupal 8';
+        if (file_exists($siteDir . '/wp-cli.yml')) {
```
greg-1-anderson: Maybe test for "$siteDir/web/wp-admin/index.php"?

stevector: The Composer-based way of doing WordPress moves `wp-admin` too. In our example it is `web/wp/wp-admin`. I don't think we can assume that. I'm leaning toward `web/wp-config.php`. @ataylorme, do you have thoughts on what file would be the best to check for to determine if a repo is WordPress?

greg-1-anderson: `web/wp-config.php` sounds like a good option to me. Worst case, we could test two locations (e.g. `web/wp/wp-admin/index.php` and `web/wp-admin/index.php`). I don't think it's a problem to put a limitation on where WordPress can be relocated to (e.g. `wp` or nothing).

If testing `wp-config.php` sidesteps this issue, even better.

src/Commands/BuildToolsCommand.php
```diff
@@ -658,9 +659,13 @@ protected function expandSourceAliases($source)
      */
     protected function autodetectUpstream($siteDir)
     {
-        return 'Empty Upstream';
-        // or 'Drupal 7' or 'WordPress'
-        // return 'Drupal 8';
+        if (file_exists($siteDir . '/web/wp-config.php')) {
```
ataylorme: @stevector and @greg-1-anderson WordPress will first look for `wp-config.php` wherever core is. If WordPress core is in a subdirectory, say `wp`, then `web/wp/wp-config.php` will take precedence over `web/wp-config.php` if it exists.

That tripped me up when first implementing WordPress is a subdirectory via Composer as the Pantheon WordPress upstream includes `wp-config.php`. When installing core from our upstream repo into a subdirectory `web/wp/wp-config.php` must be deleted.

All that being said someone _could_ have `web/wp/wp-config.php`, or wherever they have installed WordPress core, without `web/wp-config.php` and have a properly working site. I think it's okay for us to be opinionated about using `wp` for the directory name as well as having `wp-config.php` live in the we docroot.

stevector:
https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/60 Is a switch to Circle 2 that tests D8 and WordPress.

src/Commands/BuildToolsCommand.php
greg-1-anderson: A variant of this line was already committed as part of a separate PR. We could check for both `web/wp-config.php` and `web/wp/wp-config.php`, or test some file that is always in the docroot, even when core is relocated, if such a candidate exists.

greg-1-anderson: Let's continue in #60.

greg-1-anderson:
Closing in favor of #60.

src/Commands/BuildToolsCommand.php
ataylorme: I think we should use something like `find -maxdepth 2 -type f -name 'wp-config.php` rather than hard-coding directories or guessing what subdirectory will have WordPress core.

PR #37, opened 2017-05-18
Refactor repository and CI server configuration into abstract classes

greg-1-anderson:

PR #48, opened 2017-06-20
Adds missing $this in BuildToolsBase method calls

aaronbauman:
One more for https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/37


greg-1-anderson:
Obviously that last commit was a work-in-progress that I shouldn't have put in. Thanks for fixing it up.

PR #60, opened 2017-07-11
Switch to CircleCI 2.0 and add test of example-wordpress-composer

stevector:
I expect that Circle 2.0 will make it easier to test multiple example repos.

PR #69, opened 2017-07-14
Bitbucket Provider

LukasRos:
This PR adds `BitbucketProvider.php` and thus allows specifying `--git=bitbucket` on `build:project:create` command.

Two changes have been made to `ProjectCreateCommand.php`:
* Explanation of Bitbucket option with respective credentials.
* Changed order of build tasks; otherwise CircleCI API causes error (_"No main branch set"_).


greg-1-anderson:
Well done!  Perhaps we can get some tests here once #60 is merged.

I'll help refactoring other commands so that we can get bitbucket support with the full workflow.

sukottokun:
@greg-1-anderson @LukasRos and I can help with the other commands if you like. Not sure what else needs work besides env:delete:ci; do you want us to take a look at that?

greg-1-anderson:
Tests would be the next step. #60 has been merged.

LukasRos:
Test case has been added. Before they can run the Bitbucket equivalent of _pantheon-ci-bot_ has to be created and `BITBUCKET_USER` and `BITBUCKET_PASS` environment variables set in CircleCI.

ataylorme:
Thanks for the update @LukasRos! Out of curiosity do you usually work with CircleCI when using BitBucket to host a project git repository? Wondering if we could pair hosting the repo on BitBucket with [BitBucket Pipelines](https://bitbucket.org/product/features/pipelines).

greg-1-anderson:
We can use the _pantheon-ci-bot_ for this test as well.

CircleCI supports BitBucket, and I think it is a good idea to have a BitBucket + CircleCI test to confirm that this is working. It would be great to do BitBucket Pipelines as a follow-on PR with an additional test. This is not to say that this is necessarily the highest-priority bit of next work, as we also have Travis, Jenkins, and GitLab (repository and CI service) to consider. Maybe circle back to pipelines after Jenkins? I know that is a highly requested feature.

I'll see if I can get credentials set up such that these tests will run so that we can merge this.

greg-1-anderson:
Merged with 9f4884b

PR #89, opened 2017-07-25
Only delete multidevs for PRs that have at least one closed PR

greg-1-anderson:
... and no open PRs.


greg-1-anderson:
This may sometimes leave a multidev environment behind, and never delete it. This is preferable to deleting the multidev too soon, though, so I am going to merge this, and continue to improve the behavior in the future.

jodyHamilton:
Can we get a new tag released to 1.x branch soon? I'm having PR multidev environments deleted that shouldn't have been and I see that the build-tools-ci quay image is running on latest 1.x tag, which is from 2 days before this fix. Thanks!

PR #90, opened 2017-07-27
Fix script that deletes ssh keys when tests are done.

greg-1-anderson:


greg-1-anderson:
Rolled this into #171.

PR #101, opened 2017-08-31
Put our dependency installation into a docker image

greg-1-anderson:


greg-1-anderson:
I think this would work if I put our image on Dockerhub instead of quay.io.

Here are my sources:

https://github.com/pantheon-systems/docker-php-ci
https://quay.io/repository/pantheon-public/php-ci?tab=builds

Circle says you can use images from any source; I just need to figure out how to specify Quay.

stevector:
What's the advantage of Quay over dockerhub?

greg-1-anderson:
Here we are:
```
- image: quay.io/project/image:tag
```

.circleci/set-up-globals.sh
```diff
@@ -2,15 +2,6 @@
...
-# The section would be transferable to a DOCKERFILE
-apt-get update
-apt-get install -y ruby
-gem install circle-cli
-composer global require -n "hirak/prestissimo:^0.3"
-git clone https://github.com/pantheon-systems/terminus.git ~/terminus
-cd ~/terminus && composer install
-ln -s ~/terminus/bin/terminus /usr/local/bin/terminus
-
 # Commands below this line would not be transferable to a docker container
```
stevector: We can probably delete this comment too now :)

greg-1-anderson:
I am using quay.io instead of Dockerhub because @joemiller wanted it that way. 😄 

Pantheon is already using quay.io for all of our private images; I do not quite remember the things it does better that lead to that decision. Given that we already have repositories on Quay, it made sense to have our public images in the same place.  However, I made a separate repository for the public images called `pantheon-public`.

.circleci/set-up-globals.sh
greg-1-anderson: Good point!

PR #108, opened 2017-09-16
Use build:comment:add:commit in build:env:create

LukasRos:
Also renamed command.


greg-1-anderson:
Looks good.

Could you please push this to the main repository, so that the tests will run?
```
$ git remote add upstream git@github.com:pantheon-systems/terminus-build-tools-plugin.git
$ git push -u upstream build-commit-comment-command
```
You might need to create a new PR, but I think you can probably edit the branches.

Also, the `master` branch is failing after merging your last PR. Could you please take a look at that as well, and see if you can fix it?

LukasRos:
I've pushed this to the main repo. All the tests complete successfully except for *test_drupal_7_github* - this is the same behavior as on the master; I thus believe that this failing test is not related to any modified build tools command but somehow connected to Drupal 7.

The last message that the build script generates before erroring is "Collect test metadata", which I don't know where it comes from (haven't found any string like this in the build tools).

greg-1-anderson:
screen shot 2017-09-21 at 12 39 31 pm


Don't delete the --notify flag; keep it present, but ignore its contents so that existing builds will not break. We can remove --notify from existing builds over time, but upgrades / backwards compatibility will be easier if you leave it in.

You can find the tests that the terminus-build-tools-plugin tests create at https://circleci.com/gh/pantheon-ci-bot

greg-1-anderson:
+1 LGTM

PR #111, opened 2017-09-27
Multi-Git-provider support for build:env:delete

LukasRos:
Implemented support for multiple Git providers - auto detected from repo - in build:env:delete.

### DIFF
--- a/src/ServiceProviders/RepositoryProviders/Bitbucket/BitbucketProvider.php
+++ b/src/ServiceProviders/RepositoryProviders/Bitbucket/BitbucketProvider.php
@@ -207,6 +207,40 @@ class BitbucketProvider implements GitProvider, LoggerAwareInterface, Credential
         ]);
     }
 
+    /**
+     * @inheritdoc
+     */
+    function branchesForPullRequests($target_project, $state)
+    {
+        if (!in_array($state, ['open', 'closed', 'all']))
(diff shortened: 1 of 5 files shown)


src/ServiceProviders/RepositoryProviders/Bitbucket/BitbucketProvider.php
```diff
@@ -207,6 +207,40 @@ public function commentOnCommit($target_project, $commit_hash, $message)
         ]);
     }
 
+    /**
+     * @inheritdoc
+     */
+    function branchesForPullRequests($target_project, $state)
+    {
+        if (!in_array($state, ['open', 'closed', 'all']))
```
greg-1-anderson: This would be more efficiently expressed as an associative array mapping from state to parameters. Optional.

greg-1-anderson:
+1 LGTM. Do the optional refactoring if you feel like it & then merge when ready.

src/ServiceProviders/RepositoryProviders/Bitbucket/BitbucketProvider.php
LukasRos: Made a change to simplify this code and get rid of the `switch` statement. Let me know if this is what you had in mind.

PR #124, opened 2017-12-30
Updated readme with a link to the installation

carl-alberto:
I was trying to run first the build command when I was trying to follow the instructions because that was the first code that I saw and ran into an error. I have to read the whole instruction first to find out that I need to do the installation instruction at the bottom of the document, might help first time users to know the Terminus plugin installation before running any code outlined in this doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ See below for the list of supported commands. This plugin is only available for 
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks. Sorry for the enormously long delay in merging this.

PR #128, opened 2018-01-30
Fixes logic issues in branch-preservation

aaronbauman:
In my environment with zero open and zero closed pull requests, build:env:delete (and build:env:delete:ci) refuses to delete any environments.
This patch fixes the logic problem.

Works in my limited testing, but haven't tested against environments with open PRs or multidev-branches that need to be preserved.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -806,12 +806,6 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         // Add any that match "pr-BRANCH", for some BRANCH in branch-name.
         $result = array_merge($result, $this->findBranches($oldestEnvironments, array_values($closedBranchList), $multidev_delete_pattern));
 
-        // If there are no closed pull requests, then there is no need
-        // to look for open pull requests
-        if (empty($result)) {
-            return $result;
-        }
-
         $openBranchList = $this->git_provider->branchesForPullRequests($project, 'open');
 
         // Remove any that match "pr-NNN" and have an open pull request
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -794,24 +794,26 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         return preg_replace('#[^:/]*[:/]([^/:]*/[^.]*)\.git#', '\1', str_replace('https://', '', $url));
     }
 
+    /**
+     * Given an array of environments, return an array of those environments
+     * without any open pull requests
+     *
+     * @param string $remoteUrl
+     * @param array $oldestEnvironments
+     * @param string $multidev_delete_pattern
+     *
+     * @return array environmentsWithoutPRs
+     */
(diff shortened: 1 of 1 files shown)


src/Commands/EnvDeleteCommand.php
```diff
@@ -167,6 +167,10 @@ function ($item) {
         elseif (!empty($options['preserve-if-branch'])) {
             $environmentsWithoutPRs = $this->preserveEnvsWithBranches($oldestEnvironments, $multidev_delete_pattern);
         }
+        else {
+          $environmentsWithoutPRs = $oldestEnvironments;
```
greg-1-anderson: This is too destructive. delete:env should have one or the other of preserve-prs or preserve-if-branch set.

src/Commands/BuildToolsBase.php
```diff
@@ -806,12 +806,6 @@ protected function preserveEnvsWithOpenPRs($remoteUrl, $oldestEnvironments, $mul
         // Add any that match "pr-BRANCH", for some BRANCH in branch-name.
         $result = array_merge($result, $this->findBranches($oldestEnvironments, array_values($closedBranchList), $multidev_delete_pattern));
 
-        // If there are no closed pull requests, then there is no need
-        // to look for open pull requests
```
greg-1-anderson: Yes, it looks like this optimization is not correct.

src/Commands/EnvDeleteCommand.php
aaronbauman: Removed.
I think the change in BuildToolsBase is all that's needed.

aaronbauman:
Revisiting once more: preserveEnvsWithOpenPRs() shouldn't concern itself with closed PRs at all.
The purpose of this method is to return a list of branches *without* open PRs, given a list of branches. 

Thus, the only thing it needs to do is remove branches with open PRs from the given list.

Latest commit fixes this logic and adds a docblock.

aaronbauman:
are you waiting on me for something else for this one?

greg-1-anderson:
I was feeling ambivalent about how this should work, and then lost track of it. Let's just merge it like this and see how it works for folks. We'll leave it on the master branch only for a while, and then tag a release.

aaronbauman:
Understood.
I have a couple projects running through build-tools/circle now, so it has been getting a bit more annoying to manually clean out the ci-# environments.
Thanks

aaronbauman:
I guess it didn't work?
I'm still getting max-multidev issues, where they're not getting cleaned up after PRs get merged.

greg-1-anderson:
Are you using the master branch?

aaronbauman:
Yes.
Looks like this commit was reverted the day after being merged.
https://github.com/pantheon-systems/terminus-build-tools-plugin/commit/7feb606e12bdffb4112a95bfbe07c549c9e1a228

greg-1-anderson:
Hm, yes, that was a while ago; I seem to recall folks reported having multidevs deleted that should not have been deleted. Not sure; should re-investigate.

e0ipso:
> Hm, yes, that was a while ago; I seem to recall folks reported having multidevs deleted that should not have been deleted. Not sure; should re-investigate.

This is the suspected root-cause for #418

PR #130, opened 2018-02-21
BitBucket Username and Password need to be URL encoded in token.

CWCorrigan:
pushRepository() will fail for BitBucket if there is an '@' in either username or password. URL Encoded both sides of the token to escape any unwanted characters in the request.


greg-1-anderson:
LGTM thanks for this.

PR #134, opened 2018-03-19
Add CircleCI project page mapping for BitBucket repositories

NickWilde1990:
This will fix the invalid link/image to CircleCI that is in the generated readme when using a bitbucket repo


NickWilde1990:
A *very* minimal change. Would you prefer more single feature fixes as I do them or a bigger batch of multiple?

greg-1-anderson:
A small fix like this for a feature that is already working is fine; however, bitbucket support is not working yet, so it would be better to make a more complete PR with a passing test.

Since this PR is trivial, I'll go ahead and merge it.

PR #135, opened 2018-04-01
Improved bitbucket support

NickWilde1990:
This seems to work fine however, the automated tests fail for all the branches on my local. So I'm going to have to dig more into figuring that out or have you merge/pull my branch and test on your CI instance?

PR #136, opened 2018-04-01
Nick wilde1990 improved bitbucket support

greg-1-anderson:
Duplicate of #135, but with permissions to run tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,9 +58,8 @@ jobs:
                 name: dependencies
                 command: ./.circleci/set-up-globals.sh
             - run:
-                name: Test D8 with BitBucket (unfinished)
-                command: echo "Disabling this test until it is passing."
-                # command: ./.circleci/test-example-repo.sh "pantheon-systems/example-drops-8-composer:dev-test-2.x" bitbucket
+                name: Test D8 with BitBucket
+                command: ./.circleci/test-example-repo.sh "pantheon-systems/example-drops-8-composer:dev-master" bitbucket
             - run:
                 name: D8 with Bitbucket Build Failed
                 when: on_fail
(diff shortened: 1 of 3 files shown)


NickWilde1990:
hmm I dunno why those tests are failing other than the bitbucket one. I'll get my branch working I guess.

philltran:
@NickWilde1990 Thank you for your work on this. 

I was able to connect and push to bitbucket if I created a token using an oauth consumer and putting the token in the authorization header.  How are you generating a bitbucket token? I don't see the option to generate one for bitbucket cloud like github can generate a token. https://developer.atlassian.com/cloud/bitbucket/oauth-2/

philltran:
How about using an app specific password? 

> Using an app password
An app password is a substitute password for the user account where you configure it, so you simply use it when authenticating with Bitbucket:

username: your normal Bitbucket username
password: the app password.
This applies to direct API access (e.g. via curl with HTTP authentication) as well as for tools that integrate with Bitbucket via the HTTP API. As mentioned above, you cannot log in to the Bitbucket web interface with an app password.

>

philltran:
I was able to get the workflow to work using a Bitbucket app password instead of token.

Like I saw @nickwilde1990 do I had to make a custom docker image with the forked build tools for Circle CI to use when running through the jobs. 

So @nickwilde1990 gets credit, do you want me to open up a new pull request or will you edit the code on this PR?

Thanks.

greg-1-anderson:
@philltran If you send me your DockerFile, I will review it and publish it in Pantheon's Quay repository.

If you pull this branch locally and add your own commits, then if you make a new pull request, existing contributors will still appear in the git history.

philltran:
@greg-1-anderson  Thank you for the advice. I will make a new pull request. 

I traced down what was happening with the Docker Image. The docker image currently installs the 2.0.0-alpha5 version of terminus build tools. This version does not the bitbucket updates. So I will do a pull request here first. Once that is committed we will need to do a PR to change to next tag of terminus build tools.

greg-1-anderson:
Is that the only reason you needed to make a new Docker container -- to get the right version of the Terminus Build Tools plugin?  The existing tests should be run against the exact commit that's in the PR. Of course, if you are committing from outside the org, then the tests won't work (no credentials available).

philltran:
Now that you say that, it was not failing with this test here. It was failing on Circle CI during build and test jobs. I will submit my pull request for testing

PR #154, opened 2018-11-08
During Project Build Process add check to see if Terminus Session has machine token present

### LINKED ISSUE #114: Incorrect error message when not logged in
If the user is not logged in or authentication is not valid, `terminus build:project-create` displays _[Pantheon\Terminus\Exceptions\TerminusException] The site name xxx is already taken on Pantheon._ instead of a more appropriate notification error.

philltran:
Fixes #114

### DIFF
--- a/src/Commands/ProjectCreateCommand.php
+++ b/src/Commands/ProjectCreateCommand.php
@@ -88,6 +88,13 @@ class ProjectCreateCommand extends BuildToolsBase implements PublicKeyReciever
             $site_name = $target;
         }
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
I think we can rely on `@authenticated` now. Need to test to confirm.

greg-1-anderson:
Confirmed this is fixed in Terminus 2.0; we do not need to explicitly check in the Build Tools plugin now.

philltran:
Awesome. Thanks.

PR #166, opened 2019-01-25
BT-20: follow pagination links in collection responses

e0ipso:
The calls to the GitHub API may result in a collection of items. When
that happens GitHub may paginate the number of results per request.
If that is the case, there will be pagination links in the response
headers.

This PR will introspect the response to determine if it is for a
paginated resource with more results outside of the current response.
The intent is to accumulate the results for that response so the result
set is complete

### DIFF
--- a/src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
+++ b/src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
@@ -267,7 +267,6 @@ class GitHubProvider implements GitProvider, LoggerAwareInterface, CredentialCli
             $pager_info = $this->getPagerInfo($headers['Link']);
             if (!$this->isLastPage($uri, $pager_info)) {
                 $next_page_uri = $this->getNextPageUri($pager_info);
-                var_dump($next_page_uri);
                 // Request the next page and append the data.
                 $resultData = array_merge_recursive(
                     $resultData,
--- a/src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
+++ b/src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
@@ -201,7 +201,7 @@ class GitHubProvider implements GitProvider, LoggerAwareInterface, CredentialCli
         if (!in_array($state, ['open', 'closed', 'all']))
             throw new TerminusException("branchesForPullRequests - state must be one of: open, closed, all");
 
-        $data = $this->gitHubAPI("repos/$target_project/pulls?state=$state");
+        $data = $this->gitHubAPI("repos/$target_project/pulls?state=$state", [], 'GET', TRUE);
         $branchList = array_column(array_map(
             function ($item) {
                 $pr_number = $item['number'];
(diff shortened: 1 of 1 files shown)


src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
```diff
@@ -255,12 +255,106 @@ protected function gitHubAPI($uri, $data = [], $method = 'GET')
         $message = isset($resultData['message']) ? "{$resultData['message']}." : '';
 
         if (!empty($message) || !empty($errors)) {
-            throw new TerminusException('{service} error: {message} {errors}', ['service' => $service, 'message' => $message, 'errors' => implode("\n", $errors)]);
```
e0ipso: Static code analysis revealed that the `$service` variable was not defined.

src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
```diff
@@ -255,12 +255,106 @@ protected function gitHubAPI($uri, $data = [], $method = 'GET')
...
+                $resultData = array_merge_recursive(
+                    $resultData,
+                    $this->gitHubAPI($next_page_uri, $data, $method, TRUE)
+                );
+            }
+        }
         return $resultData;
     }
 
+    protected function isSequentialArray($input)
```
e0ipso: The way we parse JSON responses will always yield a PHP array. This method checks if the array comes from `{foo: 'bar'}` or from `['foo', 'bar']`. Only the latter will be considered a response to a _collection endpoint that can be paginated_.

src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
```diff
@@ -210,11 +210,11 @@ function ($item) {
             },
             $data
         ), 1, 0);
- 
+
         return $branchList;
      }
 
-    protected function gitHubAPI($uri, $data = [], $method = 'GET')
+    protected function gitHubAPI($uri, $data = [], $method = 'GET', $follow_next_link = FALSE)
```
e0ipso: This will take the cautious approach and only enable this behavior if explicitly requested. However I think this precaution may be lifted with the current code. The only limitation would be that we would no longer be capable of requesting a single page. I think requesting a single page is a useless feature, as pages are merely an implementation detail to mitigate unforeseen performance implications. Such performance degradation is preferred in a CLI application in favor of correctness.

@greg-1-anderson please let me know if you'd like me to lift this guard.

e0ipso:
See this in action here: https://asciinema.org/a/ULIdS0doYR4f51pvbwfJR9PAv

Before this PR there was only one ` [notice] Call GitHub API: GET …/pulls?state=closed`.

src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
greg-1-anderson: I'm -1 on the extra parameter. I'm undecided at this point if the feature needs to be guarded.

Sometimes you might make an API call just to get the first response. Automatically traversing all pages would not be desirable in that instance, as some paged responses, e.g. a search, could have many many pages.

greg-1-anderson: I'll look at this more later with an opinion on whether there should be a parameter here, or a separate method for paged vs unpaged access.

e0ipso:
@greg-1-anderson please let me know if you need something else from me to move this forward.

Thanks!

greg-1-anderson:
OK I'm 👍 on just leaving the extra parameter there. I just need to copy this PR branch to `origin` so that we can see the tests run.

e0ipso:
Great @greg-1-anderson thanks for looping back on this.

greg-1-anderson:
Thank you for providing this. I refactored your PR a bit and continued the work at #171.

PR #171, opened 2019-02-23
BT-20: Ensure that automatic multidev deletion is working correctly

greg-1-anderson:


src/Commands/EnvDeleteCommand.php
```diff
@@ -204,11 +185,11 @@ function ($item) {
         }
 
         // Delete each of the selected environments.
-        foreach ($oldestEnvironments as $env_id) {
+        foreach ($environmentsToDelete as $env_id) {
             $site_env_id = "{$site_id}.{$env_id}";
```
rvtraveller: I don't think `$site_id` is defined for this function. When I run this command locally I get a notice about an undefined variable and then a Terminus error about giving the environment argument properly.

greg-1-anderson:
Whoops, that was a loose end all right. I was testing with `--dry-run`, and forgot to actually exercise the code. There is no coverage for this in the functional tests yet.

rvtraveller:
LGTM

PR #175, opened 2019-02-25
BT-19: Fix terminus build:env:delete:pr for GitLab

ataylorme:


ataylorme:
Now that we are getting a valid response and not a `404` from the merge requests API call I am seeing new warnings. The response from GitLab.com and what the code expect don't align.

```
PHP Notice:  Undefined index: number in /Users/andrewtaylor/.terminus/plugins/terminus-build-tools-plugin/src/ServiceProviders/RepositoryProviders/GitLab/GitLabProvider.php on line 317

Notice: Undefined index: number in /Users/andrewtaylor/.terminus/plugins/terminus-build-tools-plugin/src/ServiceProviders/RepositoryProviders/GitLab/GitLabProvider.php on line 317
PHP Notice:  Undefined index: head in /Users/andrewtaylor/.terminus/plugins/terminus-build-tools-plugin/src/ServiceProviders/RepositoryProviders/GitLab/GitLabProvider.php on line 318

Notice: Undefined index: head in /Users/andrewtaylor/.terminus/plugins/terminus-build-tools-plugin/src/ServiceProviders/RepositoryProviders/GitLab/GitLabProvider.php on line 318
```

Here is the response from GitLab being processed:

```json
[
  {
    "id": 24756124,
    "iid": 1,
    "project_id": 11022366,
    "title": "Add Pantheon Advanced Page Cache Selective Purge Test",
    "description": "",
    "state": "merged",
    "created_at": "2019-02-25T15:54:22.640Z",
    "updated_at": "2019-02-25T17:03:12.735Z",
    "merged_by": {
      "id": 3468723,
      "name": "Andrew Taylor",
      "username": "ataylorme",
      "state": "active",
      "avatar_url": "https://secure.gravatar.com/avatar/65a38e6607bb16a7314f098d172fea38?s=80&d=identicon",
      "web_url": "https://gitlab.com/ataylorme"
    },
    "merged_at": "2019-02-25T17:03:12.760Z",
    "closed_by": null,
    "closed_at": null,
    "target_branch": "master",
    "source_branch": "papc-test",
    "upvotes": 0,
    "downvotes": 0,
    "author": {
      "id": 3468723,
      "name": "Andrew Taylor",
      "username": "ataylorme",
      "state": "active",
      "avatar_url": "https://secure.gravatar.com/avatar/65a38e6607bb16a7314f098d172fea38?s=80&d=identicon",
      "web_url": "https://gitlab.com/ataylorme"
    },
    "assignee": {
      "id": 3468723,
      "name": "Andrew Taylor",
      "username": "ataylorme",
      "state": "active",
      "avatar_url": "https://secure.gravatar.com/avatar/65a38e6607bb16a7314f098d172fea38?s=80&d=identicon",
      "web_url": "https://gitlab.com/ataylorme"
    },
    "source_project_id": 11022366,
    "target_project_id": 11022366,
    "labels": [],
    "work_in_progress": false,
    "milestone": null,
    "merge_when_pipeline_succeeds": false,
    "merge_status": "can_be_merged",
    "sha": "bd460961a89f9d0f273051c24b38d7bf6cb992d2",
    "merge_commit_sha": "2409ab71e23675a0d17098f43b84383a35fccf3b",
    "user_notes_count": 0,
    "discussion_locked": null,
    "should_remove_source_branch": null,
    "force_remove_source_branch": true,
    "web_url": "https://gitlab.com/ataylorme/ataylor-gitlab-test-13/merge_requests/1",
    "time_stats": {
      "time_estimate": 0,
      "total_time_spent": 0,
      "human_time_estimate": null,
      "human_total_time_spent": null
    },
    "squash": true,
    "approvals_before_merge": null
  }
]
```

ataylorme:
[`5f01524`](https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/175/commits/5f0152432e32d96adbbdd7112a53a1fb9d492f94) formats the output of `branchesForPullRequests` in `GitLabProvider.php` to be a valid array of `PR-Number => branch-name` that is expected.

Even with that `MultiDevRetention` is showing the closed PR on the `retain` list and not `eligible`

ataylorme:
When checking the calls to `branchesForPullRequests` it seems it is only called once with the state `all`.

greg-1-anderson:
@ataylorme The requirements for using branchesForPullRequests changed. There are open cards for follow-on work for GitLab and BitBucket. It might make sense to factor out some of the code from the GitHub API class for re-use in the other service APIs

ataylorme:
> The response from GitLab.com and what the code expect don't align

Turns out the code for `branchesForPullRequests` in `GitLabProvider.php` parsing `$branchList` was copied as-is from GitHub.

I updated it to align with the GitLab response.

greg-1-anderson:
The requirements for `branchesForPullRequests` has changed. It must be passed a callback now. It won't work right if it isn't rewritten to match the new API.

src/API/PullRequestInfo.php
```diff
@@ -5,7 +5,7 @@
 /**
  * PullRequestInfo caches info about pull requests
  */
-interface PullRequestInfo
```
greg-1-anderson: Hm, php 7.3 let me get away with this?

rvtraveller: When I try it in 7.2.15 I get:

`PHP Fatal error:  Interfaces may not include member variables in $HOME/.terminus/plugins/terminus-build-tools-plugin/src/API/PullRequestInfo.php on line 10`

rvtraveller: I see -- this causes a runtime error and not a compilation error. What happens if you try and run https://gist.github.com/rvtraveller/5a9f8baec5ef7b88edca8d0d21a706dc on 7.3?

greg-1-anderson: It's not super important -- interface is clearly wrong. I'll try later. It's also possible that I failed to commit the fix, or perhaps it is fixed only in #182

src/API/GitLab/GitLabAPI.php
```diff
@@ -16,6 +17,8 @@ class GitLabAPI extends WebAPI
 {
     const SERVICE_NAME = 'gitlab';
     const GITLAB_TOKEN = 'GITLAB_TOKEN';
+    const GITLAB_CONFIG_PATH = 'command.build.provider.git.gitlab_url';
```
greg-1-anderson: Using config for this is a good idea. The `command.build` namespace should be reserved for configuration that sets actual build tools plugin command options.  `command.build.project.create.shortcuts` also gets this wrong.  Maybe `build-tools.provider.git.gitlab.url` would be a good choice.

src/API/GitLab/GitLabAPI.php
```diff
@@ -48,6 +51,19 @@ protected function apiClient()
         );
     }
 
+    public static function determineGitLabUrl(Config $config)
+    {
+        // Start by seeing if GitLab CI has set a potential Project URL for us.
+        $potential_ci_url = getenv('CI_PROJECT_URL');
```
greg-1-anderson: If we use config, then it is not necessary to call `getenv` directly. The config item `build-tools.provider.git.gitlab.url` can already be overridden with the environment variable `TERMINUS_BUILD_TOOLS_PROVIDER_GIT_GITLAB_URL`.

src/API/GitLab/GitLabAPI.php
rvtraveller: Ok. Updated per your suggestion (and what you say makes sense).

src/API/GitLab/GitLabAPI.php
rvtraveller: Ahh, I didn't realize it pulled from environment variables. As I was refactoring this I realized it would be just as simple to set our own environment variable based on the creator's config vs trying to parse a GitLab CI variable. With the naming convention in place, we can remove some of this code.

src/ServiceProviders/CIProviders/GitLabCI/GitLabCIProvider.php
```diff
@@ -158,6 +158,10 @@ protected function setGitLabCIEnvironmentVars(CIState $ci_env)
             $data = ['key' => $key, 'value' => $value];
             $this->gitlabCIAPI($data, $gitlab_url);
         }
+
+        // Also set the GitLab URL for future use.
+        $data = ['key' => 'TERMINUS_BUILD_TOOLS_PROVIDER_GIT_GITLAB_URL', 'value' => $this->getGITLABURL()];
+        $this->gitlabCIAPI($data, $gitlab_url);
```
greg-1-anderson: You need the GitLab URL to retrieve this value, which contains the GitLab URL, right? Seems this isn't helpful, unless I'm misunderstanding something.

rvtraveller: This is particularly needed when trying to retrieve PR information for GitLab repos. When run via user (on their machine), the value comes from the Terminus config file. However, in the Build Tools docker container, that value isn't set (because you have no config file). Which means requests to the GitLab API go to gitlab.com instead of your self-hosted instance. By setting this environment variable here, it gets injected automatically into every CI build and BT picks it up as the config option.

greg-1-anderson: Ah, right, you don't need to ask the API for the value because it will be re-applied as an environment variable when your tests run. Obvious enough, but could you add one more line to the comment about that anyway?

src/API/GitLab/GitLabAPI.php
```diff
@@ -0,0 +1,129 @@
...
+            ]
+        );
+    }
+
+    public static function determineGitLabUrl(Config $config)
+    {
+        return $config->get(self::GITLAB_CONFIG_PATH, self::GITLAB_URL_DEFAULT);
+    }
+
+    public function getGITLABURL()
```
greg-1-anderson: Nit: camelcase would be better than all caps here & `setGITLABURL`

src/API/GitLab/GitLabAPITrait.php
```diff
@@ -0,0 +1,104 @@
...
+    return $repositoryEnvironment->token($key);
+  }
+
+  public function setToken($token)
+  {
+    $repositoryEnvironment = $this->getEnvironment();
+    $repositoryEnvironment->setToken($this->tokenKey(), $token);
+  }
+
+  public function getGITLABURL()
```
greg-1-anderson: CamelCase here too

src/ServiceProviders/CIProviders/GitLabCI/GitLabCIProvider.php
rvtraveller: Done.

PR #177, opened 2019-02-26
BT-12: Update documentation including GitLab instructions

greg-1-anderson:


README.md
```diff
@@ -3,78 +3,58 @@
...
-- GitHub: https://help.github.com/articles/creating-an-access-token-for-command-line-use/
-- Bitbucket: https://confluence.atlassian.com/bitbucket/app-passwords-828781300.html
-- Circle CI: https://circleci.com/account/api
-- GitLab: https://gitlab.com/profile/personal_access_tokens
-
-The GitHub token needs the "repo" and "delete repo" scopes.
+### Installing Build Tools 2.x:
+```
+mkdir -p ~/.terminus/plugins
+composer create-project -d ~/.terminus/plugins pantheon-systems/terminus-build-tools-plugin:^2
```
ataylorme: @mcdwayne is reporting this is not working. Maybe since the tagged version, e.g. `2.0.0-beta6`, is beta? I'm not sure if `—stability=dev` will help.

Perhaps we should keep this up to date with the latest beta release, e.g. `composer create-project -d ~/.terminus/plugins pantheon-systems/terminus-build-tools-plugin:2.0.0-beta6`, until we have a stable `2.x` release, then update to `terminus-build-tools-plugin:^2`

greg-1-anderson: You're right, this needs to be specified as `^2.0.0-beta6`. Keeping it up-to-date is optional. It's useful to do this as documentation (I do this in the docker container so that one can see at a glance that an older version could not have been included), but ^2.0.0-beta6 will still pick up a beta7 or later if available.

mcdwayne: Another issue to report. 
According to this doc: https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/fb04bbc755566dca8feacdcb16c35a9fc912eee4/DEVELOPMENT.md
"GITLAB_TOKEN A generated GitLab token - needs at least api and read_user scopes."

But when creating a token with both api and user_read, the token looks like this formulation:
xxxxxxxxxx-_xxxxxxxx
This causes the following error:
"GitLab authentication tokens should be 20-character strings containing only the letters a-z and digits (0-9). Please enter your token again."

Solution:
Creating a GitLab token with only API scope returns a token with the following formulation:
xxxxxxxxxxxxxxxxxxxx
Which works fine.

greg-1-anderson: read_user is not needed, but we should fix that limitation. We'll do that as a separate issue, https://github.com/pantheon-systems/terminus-build-tools-plugin/issues/178

PR #180, opened 2019-03-06
BT-13: Bitbucket pipelines. 

greg-1-anderson:
Add a 'key pair receiver' (needs both public and private keys) interface.

### DIFF
--- a/src/Commands/EnvObliterateCommand.php
+++ b/src/Commands/EnvObliterateCommand.php
@@ -51,14 +51,14 @@ class EnvObliterateCommand extends BuildToolsBase
         $this->providerManager()->validateCredentials();
 
         // Do nothing without confirmation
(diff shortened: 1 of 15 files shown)


greg-1-anderson:
Screen Shot 2019-03-29 at 3 12 41 PM

src/Commands/EnvObliterateCommand.php
```diff
@@ -51,14 +51,14 @@ public function obliterate($site_name)
         $this->providerManager()->validateCredentials();
 
         // Do nothing without confirmation
-        if (!$this->confirm('Are you sure you want to delete {site} AND its corresponding GitHub repository {url} and CircleCI configuration?', ['site' => $site->getName(), 'url' => $url])) {
+        if (!$this->confirm('Are you sure you want to delete {site} AND its corresponding GitHub repository {url} and CI configuration?', ['site' => $site->getName(), 'url' => $url])) {
```
ataylorme: If we are changing CircleCI to just CI should we also change GitHub to Git provider?

greg-1-anderson: Good idea

PR #182, opened 2019-03-16
BT-5: build:project:repair

greg-1-anderson:
Factor out PantheonProvider as a new 'site' provider. 
Add 'build:project:repair' command.

### DIFF
--- a/src/Commands/ProjectCreateCommand.php
+++ b/src/Commands/ProjectCreateCommand.php
@@ -27,43 +27,41 @@ use Pantheon\TerminusBuildTools\ServiceProviders\CIProviders\CIState;
 use Pantheon\TerminusBuildTools\ServiceProviders\ProviderEnvironment;
 use Pantheon\TerminusBuildTools\ServiceProviders\RepositoryProviders\RepositoryEnvironment;
 use Pantheon\TerminusBuildTools\ServiceProviders\CIProviders\CircleCIProvider;
-use Pantheon\TerminusBuildTools\Task\Ssh\PublicKeyReciever;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 /**
  * Project Create Command
- *
- * TODO: We could create a SiteProvider that implements PublicKeyReciever
- * rather than using $this to server that purpose.
  */
-class ProjectCreateCommand extends BuildToolsBase implements PublicKeyReciever
+class ProjectCreateCommand extends BuildToolsBase
 {
     use \Pantheon\TerminusBuildTools\Task\Ssh\Tasks;
     use \Pantheon\TerminusBuildTools\Task\CI\Tasks;
 
-    const PASSWORD_ERROR_MESSAGE="Admin password cannot contain the characters ! ; ` or $ due to a Pantheon platform limitation. Please select a new password.";
-
     /**
+     * Initialize the default value for selected options.
      * Validate requested site name before prompting for additional information.
      *
      * @hook init build:project:create
      */
-    public function validateSiteName(InputInterface $input, AnnotationData $annotationData)
+    public function initOptionValues(InputInterface $input, AnnotationData $annotationData)
     {
-        $ci_provider_class_or_alias = $input->getOption('ci');
         $git_provider_class_or_alias = $input->getOption('git');
         $target_org = $input->getOption('org');
         $site_name = $input->getOption('pantheon-site');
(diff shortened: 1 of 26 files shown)


src/Commands/ProjectCreateCommand.php
```diff
@@ -27,43 +27,41 @@
...
-    const PASSWORD_ERROR_MESSAGE="Admin password cannot contain the characters ! ; ` or $ due to a Pantheon platform limitation. Please select a new password.";
-
     /**
+     * Initialize the default value for selected options.
      * Validate requested site name before prompting for additional information.
      *
      * @hook init build:project:create
      */
-    public function validateSiteName(InputInterface $input, AnnotationData $annotationData)
+    public function initOptionValues(InputInterface $input, AnnotationData $annotationData)
```
ataylorme: Admin password is not initialized from the existing value and is prompted for each time. We should grab it from the build metadata.

src/ServiceProviders/ProviderManager.php
```diff
@@ -106,6 +110,14 @@ public function credentialManager()
         return $this->credential_manager;
     }
 
+    /**
+     * Give objects an opportunity to initialize themselves from the cli options
```
ataylorme: It would be great if users could specify which credentials they would like to attempt the repair on and use the existing values in build metadata for the other credentials.

For example, I may only wish to repair the CI provider API key/token. In the current implementation, I will have to pass all other credentials through the command line so that they are pre-populated and repair is only attempted on the CI provider token.

greg-1-anderson:
Blocked on https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/175 (at least for GitLab support)

ataylorme:
@greg-1-anderson I am still getting 

```PHP Fatal error:  Uncaught Error: Call to undefined method Pantheon\TerminusBuildTools\Credentials\CredentialRequest::validateFn() in .terminus/plugins/terminus-build-tools-plugin/src/Credentials/CredentialRequest.php:98```

I've tried updating the `setValidateFn` call in the `CredentialRequest('ADMIN_PASSWORD')` of `PantheonProvider` to the full namespace/class with no luck.

```php
$adminPasswordRequest = (new CredentialRequest('ADMIN_PASSWORD'))
            ->setInstructions(self::PASSWORD_INSTRUCTIONS)
            ->setPrompt(self::PASSWORD_PROMPT)
            ->setValidateFn(
                [
                    new \Pantheon\TerminusBuildTools\ServiceProviders\SiteProviders\PantheonProvider(),
                    'validAdminPassword'
                ]
            )
            ->setValidationErrorMessage(self::PASSWORD_ERROR_MESSAGE);
```

`setValidateFn` in `CredentialRequest` looks okay to me. Not sure where this is going wrong.

greg-1-anderson:
I'll look in a bit; sounds like missing code / bad merge

greg-1-anderson:
This is working in ad-hoc testing, but has an odd failure in Circle. Code seems to work, but 'circle watch' fails.

ataylorme:
I think there is some caching. I ran `terminus build:project:create d8 ataylor-build-tools-test-2019-03-27` before the `validateFn` fix and it failed.

I pulled the latest commit, which fixed the password issue, and ran `terminus build:project:repair ataylor-gitlab-test-2019-02-27-b` on an existing GitLab site.

I made a commit to the GitLab repo to fire off CI/CD and it failed because `TERMINUS_SITE` had the value `ataylor-build-tools-test-2019-03-27` (from the previous failed create attempt).

@greg-1-anderson for `build:project:repair` do we need to make sure all variables/credentials cached are cleared?

ataylorme:
@greg-1-anderson what if in [the `credentialKey` method of `CredentialManager`](https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/162f32164f73896fdd391b2623e3c2c5b2391478/src/Credentials/CredentialManager.php#L193:L203) we separated cache by site ID? Either instead of or in conjuction with user ID.

Either way I think we should clear the cache whenever `terminus build:project:repair` is run. If we separated cache by site ID then we could selectively clear the cache for that site rather than clearing the entire cache.

greg-1-anderson:
Cool. I'm adding a test that runs `build:project:repair` and makes a PR in only the GitHub + Circle test. I'll get that working before merging here.

PR #185, opened 2019-03-28
BUGS-2589: Fix crash with build:env:delete:pr / :ci

greg-1-anderson:


src/API/GitHub/GitHubAPI.php
```diff
@@ -59,7 +59,7 @@ protected function isPagedResponse($headers)
         return !empty($pager_headers);
     }
 
-    protected function getPagerInfo($links)
+    protected function getPagerInfo($headers)
```
greg-1-anderson: This is the bugfix line that makes line 64 work correctly.

PR #186, opened 2019-03-29
BUGS-2589: Test build:env:merge

greg-1-anderson:
Fixes related problems:
- build:env:merge does not work in some instances e.g. in the GitLab CI jobs due to fragility of the way the 'pantheon' remote is set by the push-to-Pantheon method, but needed later by the merge-to-master method.
- Multidev environments are never deleted due to a typo `empty()` instead of `!empty()`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,9 @@ jobs:
         steps:
             - run: echo "No unit tests yet."
 
+    # Note that we do more work on the Drupal 8 + GitHub test than in other permutations.
+    # We test creating a PR and merging it to master here, but omit those tests
+    # elsewhere in the interest of saving testing time.
     test_drupal_8_github:
         <<: *defaults
         steps:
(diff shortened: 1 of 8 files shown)


ataylorme:
Looks like the build failed looking for the multidev `pr-1` when the multidev created is named `ci-1`.

My guess is the workflow on CircleCI ran before CircleCI detected there was a PR. We've run into this race condition quite a bit.

Perhaps we should consider setting only build pull requests in CircleCI during `build:project:create`.

greg-1-anderson:
`ci-1` is created when the project is first created. `pr-1` is created by the pull request we make in our test, but it is removed once the "circle watch" completes. This test passes when "circle watch" fails to keep watching until the end of the sub-test run, which happens often. Sometimes it works, though, and then our test fails.

I think I just need to comment out this test until we can puzzle out a better way to do it.

PR #187, opened 2019-03-29
Add temporary environment variable detection to GitLab URL detection

rvtraveller:
Due to a bug in the way Terminus pulls environment configuration values with Robo, we (unfortunately) need to add manual detection.


greg-1-anderson:
My test merge master script is getting in the way here.

PR #188, opened 2019-03-29
BT-27: Use GIT_SSH_COMMAND for pushing to Pantheon

rvtraveller:
Let's the workflow just run and run and run without getting stuck waiting for the user to accept an SSH host key prompt.

### DIFF
--- a/.circleci/set-up-globals.sh
+++ b/.circleci/set-up-globals.sh
@@ -26,7 +26,6 @@ terminus --version
 set +ex
 terminus auth:login -n --machine-token="$TERMINUS_TOKEN"
 touch $HOME/.ssh/config
-echo -e "Host github.com gitlab.com bitbucket.com\n    StrictHostKeyChecking no" >> "$HOME/.ssh/config"
 git config --global user.email "$GIT_EMAIL"
 git config --global user.name "Circle CI"
 # Ignore file permissions.
--- a/.circleci/set-up-globals.sh
+++ b/.circleci/set-up-globals.sh
@@ -26,6 +26,7 @@ terminus --version
 set +ex
 terminus auth:login -n --machine-token="$TERMINUS_TOKEN"
 touch $HOME/.ssh/config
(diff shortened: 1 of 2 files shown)


ataylorme:
We should also consider disabling strict host checking for specific domains, e.g. `*.drush.in`, instead of globally.

rvtraveller:
New day yields new thoughts on how to tackle this and the revelation as to why `putenv()` wasn't working. I like this implementation much more than my previous one.

By using `putenv()`, the environment variable should only be set for the execution of the PHP code and not persisted for the user permanently. This has the advantage of not worrying about overwriting a user's preset `GIT_SSH_COMMAND` (which the prior implementation would do). It applies itself to any Build Tools command as long as it inherits from `BuildToolsBase`.

I think explicitly whitelisting certain domains is going to get tricky, especially with the introduction of the possibility of _any_ domain for GitLab.

greg-1-anderson:
So let's puzzle this through a bit more. We want to disable strict key checking to *.drush.in because those domains change dynamically, and there is no good way for the user to know whether or not they should accept the warning, so they just have to accept everything anyway. If the user is pushing to some custom GitLab domain, though, they have probably pushed to it before, and therefore have probably already accepted it as a valid known host. If they haven't, then they probably want to see the warning to protect against MITM attacks.

So, I think @ataylorme is right here -- we only want to disable checking for `*.drush.in`.

rvtraveller:
Thinking through it more, I agree with your puzzled thinking. I'll tweak.

rvtraveller:
Digging into this more, the command line options you can pass to ssh (and `GIT_SSH_COMMAND`) don't allow for you to specify an explicit host for Strict Host Key Checking. In order to only have this apply to `*.drush.in` hosts we would need to move this from an init hook back into the `pushCodeToPantheon()` function and would need to unset our variable once complete. Before I switch this up, any thoughts from anyone?

greg-1-anderson:
I think I'd rather keep it like this rather than write / remove ssh config files.

rvtraveller:
Also updated our rsync implementation to ignore SSH host keys so users aren't prompted randomly through PR deletes and such.

PR #189, opened 2019-03-30
BT-36: Fix another problem with merge to master

greg-1-anderson:
Declare namespace of PullRequestInfo before using it. Remove unused method createRepositoryInfo


aaronbauman:
Need same fix in bitbucket provider.
Possibly Gitlab too?

See https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/212

PR #190, opened 2019-03-30
Support for custom build asset messages.

### LINKED ISSUE #161: Add option for `build:env:push` command to include message
Currently, as per [this line](https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/master/src/Commands/BuildToolsBase.php#L767), deploys are all marked as "Build assets for `$env_label`" which isn't too helpful contextually in the Pantheon dashboard. It would be awesome if we were able to optionally pass a `--message` param to the push command.

First I'm just gauging interest. I'd be happy to contribute if pointed in the right direction.

rvtraveller:
Resolves #161 

By default, Build Tools will create an additional commit when pushing to Pantheon with a commit message of "Build assets for <environment>". Prior commit history from the external repository provider is retained as the "Build assets" command is just an additional commit.

Using `build:env:create` or `build:env:push` you can now include an optional `--message` flag to override the commit message on the build commit.

image

Should the `--message` option not be used, the existing functionality will be used including the "copying" of commit messages from the external provider to Pantheon and the "Build assets for <environment>"

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -701,7 +701,8 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $site_env_id,
         $multidev = '',
         $repositoryDir = '',
-        $label = '')
+        $label = '',
+        $message = '')
     {
         list($site, $env) = $this->getSiteEnv($site_env_id);
         $env_id = $env->getName();
@@ -710,6 +711,10 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $env_label = $multidev;
         if (!empty($label)) {
             $env_label = $label;
+        }
+
+        if (empty($message)) {
+            $message = "Build assets for $env_label.";
         }
 
         if (empty($repositoryDir)) {
@@ -778,7 +783,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $this->passthru("git -C $repositoryDir add --force -A .");
 
         // Now that everything is ready, commit the build artifacts.
-        $this->passthru("git -C $repositoryDir commit -q -m 'Build assets for $env_label.'");
+        $this->passthru("git -C $repositoryDir commit -q -m '$message'");
 
         // If the environment does exist, then we need to be in git mode
         // to push the branch up to the existing multidev site.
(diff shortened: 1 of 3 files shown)


src/Commands/BuildToolsBase.php
```diff
@@ -778,7 +783,7 @@ public function pushCodeToPantheon(
         $this->passthru("git -C $repositoryDir add --force -A .");
 
         // Now that everything is ready, commit the build artifacts.
-        $this->passthru("git -C $repositoryDir commit -q -m 'Build assets for $env_label.'");
+        $this->passthru("git -C $repositoryDir commit -q -m '$message'");
```
greg-1-anderson: Need to appropriately escape $message in case it contains `'`

rvtraveller: Good call.

src/Commands/BuildToolsBase.php
```diff
@@ -783,7 +783,7 @@ public function pushCodeToPantheon(
         $this->passthru("git -C $repositoryDir add --force -A .");
 
         // Now that everything is ready, commit the build artifacts.
-        $this->passthru("git -C $repositoryDir commit -q -m '$message'");
+        $this->passthru($this->interpolate("git -C {repositoryDir} commit -q -m [[message]]", ['repositoryDir' => $repositoryDir, 'message' => $message]));
```
greg-1-anderson: `repositoryDir` is using a different interpolate convention than `message`. It also looks like you dropped the quotes around $message.

rvtraveller: Right, I made the assumption that `repositoryDir` was safe so used the raw value format instead of the shell-escaped value. I think this assumption is likely fine...and if it isn't, we have a much larger task of updating the rest of Build Tools to properly escape it.

`ProcessUtils::escapeArgument()` automatically adds the quotes for `$message` -- https://github.com/symfony/process/blob/3.4/ProcessUtils.php#L78.

ataylorme:
A larger issue I see is that the commit messages from eternal providers are not sent to Pantheon. 

Can we leverage `build:env:push --message=` to pass commit messages from the external repo to the Pantheon repo?

rvtraveller:
@ataylorme can you point me to a workflow that results in what you're seeing? I've just retested GitLab and GitHub right now and in both of them commit messages (and commits) are transferring over from the external provider to Pantheon. I can ping you example sites from my side that are working as (I think) we intend if you want.

ataylorme:
@rvtraveller I will test again. It is a historic issue that may have been solved already.

PR #192, opened 2019-03-31
Include additional info in build metadata

rvtraveller:
This should, hopefully, make a few things easier for quicksilver pushback.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -996,14 +996,41 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
     }
 
     /**
+     * Sets build info from a project created by build:env:create.
+     */
+    public function setProjectInfo($url, $ci='')
+    {
+        // Create a git repository service provider appropriate to the URL
+        $this->git_provider = $this->inferGitProviderFromUrl($url);
+
+        // Extract just the project id from the URL
+        $target_project = $this->projectFromRemoteUrl($url);
+        $this->git_provider->getEnvironment()->setProjectId($target_project);
+
+        $ci_provider_class_or_alias = $this->selectCIProvider($this->git_provider->getServiceName());
+
+        $this->ci_provider = $this->createCIProvider($ci_provider_class_or_alias, $ci);
+
+        return $target_project;
+    }
+
+    /**
      * Return the metadata for this build.
      *
      * @return string[]
      */
     public function getBuildMetadata($repositoryDir)
     {
+        $url = exec("git -C $repositoryDir config --get remote.origin.url");
+        if (is_null($this->git_provider) || is_null($this->ci_provider))
+        {
+            // TODO: This doesn't account for people who specify a different CI provider (bitbucket).
+            // Should we try and retrieve build metadata from the dev site and use it?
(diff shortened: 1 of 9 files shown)


src/Commands/BuildToolsBase.php
```diff
@@ -1003,12 +1003,15 @@ protected function getLatestWorkflow($site)
...
-          'build-date'  => date("Y-m-d H:i:s O"),
+          'url'           => exec("git -C $repositoryDir config --get remote.origin.url"),
+          'ref'           => exec("git -C $repositoryDir rev-parse --abbrev-ref HEAD"),
+          'sha'           => $this->getHeadCommit($repositoryDir),
+          'comment'       => exec("git -C $repositoryDir log --pretty=format:%s -1"),
+          'commit-date'   => exec("git -C $repositoryDir show -s --format=%ci HEAD"),
+          'build-date'    => date("Y-m-d H:i:s O"),
+          'git-provider'  => $this->git_provider->getServiceName(),
+          'ci-provider'   => $this->ci_provider->getServiceName(),
+          'api-host-host' => $this->git_provider->api()->getApiHost(),
```
greg-1-anderson: Presume this should just be `api-host` and not `api-host-host`?

I'm unsure whether we should be adding this here. It is a well-known value. It seems unlikely that a script could use it without knowing which service it is for (agnostic of service, that is), as the API calls are going to vary by service. Rather than provide this constant, it might be better to provide a small library sites could link to if they needed to make an API call (probably as a future task).

rvtraveller: I think the primary purpose of putting it here was to handle those instances where people had non-standard URLs (for example, self-hosted GitLab). While we could extract it out of the Git URL for that particular service, it seemed like we would be duplicating code & effort when that time came. It didn't seem to introduce a ton of extra overhead to add it here and it means it can be used in the future without worry.

greg-1-anderson: Ah, for GitLab custom API servers that makes sense.

greg-1-anderson:
The test failure is an actual failure. See the [failing circle test](https://circleci.com/gh/pantheon-ci-bot/build-tools-1963/9?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

```
 [notice] Pushing code to pr-1 using branch pr-1.
 [notice] Running git -C /home/tester/example_drops_8_composer remote add pantheon ssh://codeserver.dev.14dd90df-b9a9-4d3f-b6a4-0c04c4d4b7ee@codeserver.dev.14dd90df-b9a9-4d3f-b6a4-0c04c4d4b7ee.drush.in:2222/~/repository.git

Fatal error: Uncaught Error: Call to a member function getServiceName() on null in /usr/local/share/terminus-plugins/terminus-build-tools-plugin/src/Commands/BuildToolsBase.php:1012
Stack trace:
#0 /usr/local/share/terminus-plugins/terminus-build-tools-plugin/src/Commands/BuildToolsBase.php(743): Pantheon\TerminusBuildTools\Commands\BuildToolsBase->getBuildMetadata('/home/tester/ex...')```

rvtraveller:
Going to take a different approach. Abandoning this.

PR #193, opened 2019-04-02
BT-30: Support for a build-providers.json file

rvtraveller:
To make Quicksilver Pushback easier/more complete we'll make a `build-providers.json` file that can include the name of the Git and CI provider as well as any other (non-sensitive) information that needs to be shared (such as API hostname).

Still to do:
- [x] Set the `git-secrets.json` file similar to how `terminus secret:set` would
- [x] Determine if `build:project:repair` should support creating this file. It currently doesn't perform an operation that involves cloning the repository.
- [x] Test


.circleci/test-secrets.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Confirm that we can retrieve a value
+singlevalue=$(terminus build:secrets:show $TERMINUS_SITE author --file=build-testing.json)
+
+test singlevalue = "rvtraveller"
+
+terminus build:secrets:delete $TERMINUS_SITE key --file=build-testing.json
+terminus build:secrets:delete $TERMINUS_SITE author --file=build-testing.json
+
+emptyvalues=$(terminus biold:secrets:list $TERMINUS_SITE --file=build-testing.json)
```
greg-1-anderson: Typo `biold`

rvtraveller:
Tests are running since master didn't merge in cleanly but this should be ready for review. It includes a test to confirm the secret commands are functional. Manual testing is always worthwhile.

`build:project:repair` will set the secrets file for the user but will not set the `build-providers.json` file in the external repository. I think `build:project:repair` should be scoped to only repairing things that *don't* involve changes to the actual repository code itself. We could consider creating something like a `build:project:convert` command that is more git focused in terms of converting an existing external repository to fit the BT workflow which I think would satisfy this need, should it ever come up.

src/Commands/BuildToolsBase.php
```diff
@@ -1227,6 +1234,59 @@ protected function exec($command)
...
+     */
+    protected function downloadSecrets($site_env_id, $filename)
+    {
+        $workdir = $this->tempdir();
+        $this->rsync($site_env_id, ":files/private/$filename", $workdir, true);
+
+        if (file_exists("$workdir/$filename"))
+        {
+            $secrets = file_get_contents("$workdir/$filename");
+            $secretValues = (array)json_decode($secrets);
```
greg-1-anderson: Use `json_decode($secrets, true);`. Passing `true` will make json_decode convert all elements in the structure to arrays. Casting to `(array)` only does the top level, and leaves everything else stdclass.

src/Commands/BuildToolsBase.php
```diff
@@ -1227,6 +1234,59 @@ protected function exec($command)
...
+    }
+
+    protected function writeSecrets($site_env_id, $secretValues, $clear, $file)
+    {
+        if (!$clear)
+        {
+            $values = $this->downloadSecrets($site_env_id, $file);
+        }
+
+        $values = array_replace($values, $secretValues);
```
greg-1-anderson: `$values` not always defined here e.g. if $clear is true

rvtraveller:
Updated to put tokens in a folder.

src/Commands/ProjectCreateCommand.php
```diff
@@ -351,6 +352,20 @@ function ($state) use ($siteDir, $source) {
...
+            ->taskPushbackSetup()
+                ->dir($siteDir)
+                ->provider($this->git_provider, $this->ci_provider)
+            ->progressmessage('Set build secrets')
+            ->addCode(
+                function ($state) use ($site_name) {
+                    $secretValues = [
+                        'token' => $this->git_provider->token($this->git_provider->tokenKey())
+                    ];
+                    $this->writeSecrets("{$site_name}.dev", $secretValues, false, '.build-secrets/tokens.json');
```
greg-1-anderson: `.build-secrets` should be enforced in the `secrets` API methods, not inserted by the caller.

src/Commands/ProjectRepairCommand.php
```diff
@@ -158,7 +158,16 @@ public function repair(
...
                 ->provider($this->ci_provider)
-                ->environment($ci_env);
+                ->environment($ci_env)
+
+            ->addCode(
+                function ($state) use ($site_name) {
+                    $secretValues = [
+                        'token' => $this->git_provider->token($this->git_provider->tokenKey())
+                    ];
+                    $this->writeSecrets("{$site_name}.dev", $secretValues, false, 'buildtools-secrets.json');
```
greg-1-anderson: Make sure this is consistent with `build:project:create`

src/Commands/SecretListCommand.php
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @command build:secrets:list
+     *
+     * @param string $site_env_id Name of the environment to set the secret in.
+     * @option file Name of file to store secrets in
+     */
+    public function delete(
+        $site_env_id,
+        $options = [
+            'file' => 'secrets.json',
```
greg-1-anderson: The default value for `file` for all of these commands should be the same as the filename used by `build:project:create` and `build:project:repair`

rvtraveller: Updated

src/Commands/ProjectRepairCommand.php
rvtraveller: Done

src/Commands/ProjectCreateCommand.php
rvtraveller: Got it.

PR #200, opened 2019-04-30
Allow git providers to override build metadata.

rvtraveller:
When GitLab CI runs a build it checks out the code for that build in a detached HEAD status. This means that when the code is pushed to Pantheon the `build-metadata.json` file reports that `'ref' => 'HEAD'` and other scripts treat this as `master` (which it isn't). There isn't a very clean way (or any way) to modify the git command as `git branch` returns `HEAD` and `git rev-parse --abbrev-ref HEAD` also yields something other than the branch name.

Recognizing that `git_provider` is not always initialized, this attempts to allow git providers to override build metadata if they exist. If not, the defaults previously used are still used. This allows the GitLab provider to use GitLab CI environment variables (if running in CI) to properly populate the branch name.


greg-1-anderson:
Maybe instead of providing a general `get` method, an `alter` method might be of more utility, at least via-a-vis flexibility for future unknown use cases.

rvtraveller:
@greg-1-anderson can you let me know if this is what you had in mind?

PR #202, opened 2019-05-07
WIP - Option to omit --force param

stevector:
I'm working on a CircleCI Orb that should first be beneficial for the use case of having the same repo on GitHub and Pantheon (without a build step). Using the `--force` option on git commands seems out of step with that use case.

I'm making this branch so that I can pull it in to some of my personal repos. If it goes well I will add a cli option to toggle `--force` off.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -815,7 +815,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
 
         // Push the branch to Pantheon
         $preCommitTime = time();
-        $this->passthru("git -C $repositoryDir push --force -q pantheon $branch");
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
`--force` is used because Build Tools overwrites the build assets every time it pushes, so the Pantheon repo is always the canonical repo + one commit with the build assets.

If you do not have a build step, or even if you do have a build step but you are maintaining the full repository contents in both repos, then it would be better to not use `--force`.

The only reservation I have is that it would be bad if some confusion caused the `--force` flag to be used when it wasn't needed (which would result in a system that worked, but that was unnecessarily dangerous), or if the `--force` flag is not used when it is needed (resulting in a failed push). I'm not sure that we could detect and do the right thing automatically though.

PR #207, opened 2019-06-03
Build assets on project creation if command exists

### LINKED ISSUE #54: Call build-assets in build:project:create
I had to back out #46, because it was calling build-assets at the wrong time in the process, causing assets to be committed to the GitHub repository.

We'll have to think of another way to do this for sites that have assets that need to be built and included in the first push to Pantheon.

NickWilde1990:
Resolves #54 
Had need for this and didn't notice the issue already existed or I would've pinged you first Greg, but anyways...


NickWilde1990:
Fails are due to circle CI not having the right access on forks it looks like to me

greg-1-anderson:
Yeah it's hard to have tests run for community-provided PRs for this project.

PR #208, opened 2019-06-03
WIP Fix a drush permission error on some systems

NickDickinsonWilde:
**This definitely shouldn't be committed currently**

On at least some systems, committing Drush as a site local dependency causes permissions errors subsequently.
Under most situations this is a minor annoyance at most - just change the permission and commit. However, with build tools, it causes a fatal error.
```
 [notice]  - drush site-install standard --account-mail='[REDACTED]' --account-name=[REDACTED] --account-pass="[REDACTED]" --site-mail='[REDACTED]' --site-name=[REDACTED] --yes
Warning: Permanently added '[appserver.dev.[REDACTED].drush.in]:2222,[35.222.201.61]:2222' (RSA) to the list of known hosts.
sh: /srv/bindings/4be17dd1a93a42318365d6aba39c8692/code/vendor/drush/drush/drush.launcher: Permission denied
 [CompletionWrapper]  Install site failed with exit code 126
 [CompletionWrapper]  Exit code 1
```

The attached PR fixes that.

However, I'm not very keen on that due to incompatibility with different systems/WP vs Drupal etc. So this mostly saying yeah there is a potential problem here.


greg-1-anderson:
Build tools shouldn't be modifying things in the vendor directory. Drush should already be +x in git. Any idea why it might not be?

nickwilde1990:
Didn't mean to flag this back, was just updating my branch for personal reasons.

I _think_ I've tracked it down to some funnyness in Git on Windows Subsystem Linux. The file shows +x within that copy of Ubuntu, but for some reason git doesn't add it as such - but once set, does maintain it. It is unrelated to Build Tools, but hard to get around without messing with Build Tools, since there isn't really a pause to manually change it.

PR #223, opened 2019-07-05
Add support for Bitbucket OAuth

aaronbauman:
I've been running into throttling issues using Bitbucket username/password auth (or username / app password auth), leading to intermittent 403 messages.

This is a proof of concept for Bitbucket OAuth Client Credentials Grant, via https://developer.atlassian.com/bitbucket/api/2/reference/meta/authentication#oauth-2


aaronbauman:
Composer supports bitbucket OAuth via its COMPOSER_AUTH / auth.json feature.

Would also be interesting to explore re-using the same token for build tools auth.

ataylorme:
@aaronbauman it looks like you are using the existing `BITBUCKET_USER` and `BITBUCKET_PASS` for the OAuth consumer key and secret. That should be fine but do you mind adding some code comments to indicate that is what is happening?

We are working through GitLab support now then plan to come back and dig into BitBucket. Once we finalize a second provider adding a third will be easier than trying to sync potential architectural changes across two new provider implementations at once.

aaronbauman:
Re-reviewing this now, I think overloading the USER and PASS paramaters is not the best approach.

For anyone relying on old fashioned auth, they'd have to wait for a failed OAuth before every bitbucket operation.

A better approach would be to create a separate API provider to use an alternative authentication scheme.

(Also, I seem to have worked around the 403 issues I was having by switching to app passwords, so this is less pressing for my own use case.)

ataylorme:
Thanks for circling back to this @aaronbauman. Glad to hear you worked around your issue. I'll leave this open for now to see if others have interest in OAuth instead of using an app password. If there isn't interest, then I don't think refactoring this will be worth the effort.

Other providers, like GitHub, require a token. I'm wondering if we should get rif of the username/password option for BitBucket and make app password the only way to go.

aaronbauman:
Bitbucket works a bit differently than Github.
There's no way to differentiate between username / password and username / app password.
So, I don't think there's any reliable way to require an app password.

ataylorme:
Gotcha. We could just update the context hints to all say app password, link to the documentation for creating an app password, etc. to recommend one.

PR #224, opened 2019-07-06
allow build:env:create to clone test, live, etc

clayfreeman:
This commit fixes an error with `build:env:create` which currently depends on the source environment being the `dev` environment (since it is the only one that has a `git_url`).

This is fixed by always using the `dev` environment for `BuildToolsBase ::addPantheonRemote()` since that is the only logical entry point to the codebase of a site. A new branch will still be created for the multidev environment as expected.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -726,6 +726,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $message = '')
     {
         list($site, $env) = $this->getSiteEnv($site_env_id);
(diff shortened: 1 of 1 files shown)


stevector:
+1, thanks for the PR @clayfreeman! @greg-1-anderson, any objection to merging this? I put a similar change in the build tools branch used by the CircleCI orb. https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/202/files

clayfreeman:
Any updates on this? Would love to get this merged & released.

stevector:
Thanks again @clayfreeman!

PR #226, opened 2019-07-17
Add functionality and a flag to allow github private repositories.

traverus:
Hello,

I've added some functionality to allow for a flag that allows for github repositories created automatically by the tool to also be private.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ Additional options are available to further customize the build:project:create c
 | --admin-email    | The email address to use for the admin |
 | --ci             | The CI provider to use. Defaults to "circleci" |
 | --git            | The git repository provider to use. Defaults to "github" |
+| --visibility     | The visibility of the project. Defaults to "public". Use "public" or "private" for GitHub and "public", "private", or "internal" for GitLab |
 
 See `terminus help build:project:create` for more information.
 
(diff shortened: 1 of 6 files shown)


ataylorme:
@traverus thanks for starting this PR! We are working on adding GitLab support so I think this can be generalized to a flag for `visibility`, which defaults to `public`, instead of `github-private`.

We will probably have to wait for the GitLab work to land to figure out the best approach to make this more generic.

rvtraveller:
I updated this to also include support for GitLab. Before we merge it, however, I have 1 concern we'll need to resolve (not counting anything @ataylorme has or similar).

This implementation, in its current form, maintains consistency in functionality between repository providers. In other words, projects created in GitHub and GitLab have the same set of defaults (being created as public). However, in my interactions with GitLab projects default to `private` for me when created via API. Are we willing to accept variation between repository providers such that GitHub defaults to "public" but GitLab defaults to "private"?

ataylorme:
>  Are we willing to accept variation between repository providers such that GitHub defaults to `public` but GitLab defaults to `private`?

In my opinion, we should not. This would be confusing. I would much rather have the same set of default for Build Tools, even those defaults differ from the repository providers defaults.

ataylorme:
Tests are running is #233

PR #227, opened 2019-07-18
Gracefully fail if build:env:create source and multidev are the same

davereid:


ataylorme:
Running tests in #235

PR #228, opened 2019-08-02
WIP: Add testing for self-hosted GitLab

rvtraveller:


.circleci/config.yml
```diff
@@ -167,6 +167,52 @@ jobs:
...
+              - run:
+                  name: Check Terminus version
+                  command: terminus --version
+              - run:
+                  name: Test D8 with GitLab
+                  command: ./.circleci/test-gitlab-repo.sh "pantheon-systems/example-drops-8-composer:dev-master" gitlabci
+              - run:
+                    name: Watch GitLab CI tests
+                    command: ./.circleci/watch-gitlabci.sh
+              - run:
```
ataylorme: Instead of a pass and fail step we should use a singular step with `when` set to `always`

Issue #199, opened 2019-04-24
Fatal when creating a project from bitbucket in version 2.x; tokenKey() method missing


dkosbob:
ProjectCreateCommand->createProject calls tokenKey() from the git_provider, which results in a fatal error when the provider is bitbucket.

There should be a tokenKey() method in BitbucketAPITrait as there is in GitLabAPITrait and GitHubAPITrait.

vinmassaro:
I am also running into this using `terminus build:project:create --git=bitbucket`.

ataylorme:
The BitBucket work is still in progress and very much incomplete. You can use it but until Terminus Build Tools 2.0 launches BitBucket support is unstable.

Are either of you interested in contributing to the development of BitBucket support?

antonioG4:
@ataylorme I'm taking a stab at this, can you provide me proper github permissions?

ataylorme:
> I'm taking a stab at this, can you provide me proper github permissions?

@antonioG4 that's great! Unfortunately, Pantheon policy doesn't allow for giving write permissions to our repositories without an MSA for security reasons. CircleCI jobs, for example, mirrors GitHub permissions and allows jobs to be run with SSH access, potentially exposing authentication credentials.

I think the best approach would be to fork the build tools repository, do your work there, and then submit a pull request. Aside from using our example repositories as a start state for new build tools projects you can specify a custom repository. That will allow you to fork [`example-wordpress-composer`](https://github.com/pantheon-systems/example-wordpress-composer/) or [`example-drops-8-composer`](https://github.com/pantheon-systems/example-drops-8-composer/) as well and add a BitBucket config file to them.

We've tried to keep the Bash scripts in those repositories provider agnostic and use more generic environment variables. [The `.gitlab-ci.yml` in `example-wordpress-composer`](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/.gitlab-ci.yml) is a good example.

For the build tools work, the code base is pretty abstracted. If it would be helpful we can try to schedule a call with @greg-1-anderson to walk through it so you know where the provider-specific integration points are.

If there is anything else we can do to help, let me know.

antonioG4:
@ataylorme Done.  The simple adding of the token method appears to have worked in my case.  However, I think I need to repeat on a completely clean env, as I wasn't prompted for credentials again.

ataylorme:
[...]

PR #229, opened 2019-08-05
#199 - add tokenKey method for bitbucket trait

antonioG4:
I added the tokenKey method, and linked the properties required.

### DIFF
--- a/src/API/Bitbucket/BitbucketAPITrait.php
+++ b/src/API/Bitbucket/BitbucketAPITrait.php
@@ -45,6 +45,11 @@ trait BitbucketAPITrait
         return $api;
     }
 
+    public function tokenKey()
+    {
+        return BitbucketAPI::BITBUCKET_PASS;
+    }
+
     public function hasToken($key = false)
     {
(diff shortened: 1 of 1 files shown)


ataylorme:
@antonioG4 thank for the PR! Do you have a project source repo with BitBucket config for testing? If not let's work together to create one 😄

src/API/Bitbucket/BitbucketAPITrait.php
```diff
@@ -45,6 +45,11 @@ protected function createApi($environment)
         return $api;
     }
 
+    public function tokenKey()
```
ataylorme: We should update `setToken` to use the new method as well.

ataylorme:
This is required for Quicksilver Pushback

src/API/Bitbucket/BitbucketAPITrait.php
ataylorme: Nevermind, `setToken` uses `BitbucketAPI::BITBUCKET_AUTH`.

ataylorme:
Running tests in #234

PR #243, opened 2019-08-28
Documentation updates

rvtraveller:
Improve README documentation to include available commands and other updates based on all work done to date.


ataylorme:
@rvtraveller can you add documentation on self-hosted GitLab?

PR #248, opened 2019-09-01
Revert to build:env:push if build:env:create is run against dev

### LINKED ISSUE #247: build:env:create fails when the multidev option is dev
Instead, `build:env:create` should revert to `build:env:push` functionality. See https://github.com/pantheon-systems/docker-build-tools-ci/issues/35.

Even though `dev` is not a valid multidev environment this worked prior to #225 and breaking existing builds is not something we want to do.

ataylorme:
Closes #247

### DIFF
--- a/src/Commands/EnvCreateCommand.php
+++ b/src/Commands/EnvCreateCommand.php
@@ -59,11 +59,8 @@ class EnvCreateCommand extends BuildToolsBase
         // Revert to build:env:push if build:env:create is run against dev.
         if ('dev' === $multidev) {
             $this->log()->notice('dev has been passed to the multidev option. Reverting to dev:env:push as dev is not a multidev environment.');
-            // Get the site ID
-            $siteInfo = $site->serialize();
-            $site_id = $siteInfo['id'];
-            // Run build:env:push. Ignore errors.
-            passthru('terminus build:env:push ' . $site_id . '.dev');
+            // Run build:env:push.
+            $this->pushCodeToPantheon($site_env_id, $multidev, '', $env_label);
             return;
         }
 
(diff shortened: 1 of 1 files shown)


ataylorme:
@sean-e-dietrich can you review/test?

src/Commands/EnvCreateCommand.php
```diff
@@ -56,8 +56,19 @@ public function createBuildEnv(
...
     {
         list($site, $env) = $this->getSiteEnv($site_env_id);
+        // Revert to build:env:push if build:env:create is run against dev.
+        if ('dev' === $multidev) {
+            $this->log()->notice('dev has been passed to the multidev option. Reverting to dev:env:push as dev is not a multidev environment.');
+            // Get the site ID
+            $siteInfo = $site->serialize();
+            $site_id = $siteInfo['id'];
+            // Run build:env:push. Ignore errors.
+            passthru('terminus build:env:push ' . $site_id . '.dev');
```
greg-1-anderson: Couldn't we call through to the appropriate API rather than use 'passthru' here?

greg-1-anderson: Further down in this command we call pushCodeToPantheon. Exit after that call if we're in the `dev` environment. Make sure we don't exit early if creating dev from dev. Make sure that the code realizes the multidev exists if it's the dev environment (maybe no code change needed).

That's two or three steps instead of one, but worth it to avoid the passthru.

ataylorme: @greg-1-anderson I updated to use the `pushCodeToPantheon` method instead of `passthru` but am still doing that early and exiting afterward.

I don't see the benefit in adding more steps to the logic before the existing `pushCodeToPantheon` call instead of just calling `pushCodeToPantheon` early.

PR #268, opened 2019-09-11
Allow git providers to provide build secrets.

aaronbauman:
Allowing git providers to specify their own build secrets addresses issues like https://github.com/pantheon-systems/quicksilver-pushback/issues/12 where different providers expect different credentials.

Adds `GitProvider::getSecretValues` interface method, and default implementation to match existing behavior.

### DIFF
--- a/src/Commands/ProjectRepairCommand.php
+++ b/src/Commands/ProjectRepairCommand.php
@@ -162,10 +162,8 @@ class ProjectRepairCommand extends BuildToolsBase
 
             ->addCode(
                 function ($state) use ($site_name) {
-                    $secretValues = [
-                        'token' => $this->git_provider->token($this->git_provider->tokenKey())
-                    ];
-                    $this->writeSecrets("{$site_name}.dev", $secretValues, false, 'tokens.json');
+                  $secretValues = $this->git_provider->getSecretValues();
+                  $this->writeSecrets("{$site_name}.dev", $secretValues, false, 'tokens.json');
                 }
             );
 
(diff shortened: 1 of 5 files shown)


src/Commands/ProjectRepairCommand.php
```diff
@@ -162,10 +162,8 @@ public function repair(
 
             ->addCode(
                 function ($state) use ($site_name) {
-                    $secretValues = [
-                        'token' => $this->git_provider->token($this->git_provider->tokenKey())
-                    ];
-                    $this->writeSecrets("{$site_name}.dev", $secretValues, false, 'tokens.json');
+                  $secretValues = $this->git_provider->getSecretValues();
+                  $this->writeSecrets("{$site_name}.dev", $secretValues, false, 'tokens.json');
```
ataylorme: @aaronbauman won't `build:project:repair` miss the provider specific additions to `tokens.json`? `getSecretValues` only returns `token`.

ataylorme: Ah, wait nevermind, it is calling `getSecretValues` for the provider, not the method from the base class

aaronbauman: `BitbucketProvider::getSecretValues()` returns `user` and `password` additionally, so should work.

https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/268/files#diff-a39100c401681ff242854edb6766562aR147

ataylorme:
Tests are running on #269

ataylorme:
> Tests are running on #269

Nevermind, you have access so the tests are running here

ataylorme:
I tested this manually with `build:project:create` and BitBucket. `tokens.json` had entries for `token`, `user` and `pass`.

Quicksilver pushback still fails because of the `url` value in `build-metadata.json` starting with `http://bitbucket.org` and not `https` but that is out of scope for this PR.

Issue #275, opened 2019-09-16
Support for private packages


aaronbauman:
I've got a composer base project downstream of https://github.com/pantheon-systems/example-drops-8-composer/ that relies on private packages.
Once I have all my auth in place, everything works great.

However, the first build for my initial spinup fails because of my reliance on private packages. I'm working around this by creating a `COMPOSER_AUTH` variable immediately after `build:project:create` - which is fine, but makes the process more manual than I'd like.

Is there a way I could be doing this to avoid the manual step?

I'm assuming no, but would welcome help in formulating a PR to support this workflow.

ataylorme:
@aaronbauman does exporting `COMPOSER_AUTH` _before_ running `build:project:create` work for you?

Build Tools could potentially do this and store the Composer auth JSON value the same way it stores other auth credentials, such as `BITBUCKET_PASS`.

ataylorme:
In scope:
- Ensue if `COMPOSER_AUTH` is a defined environment variable, it is passed through to the `build:project:create` process
- Allow for reading `COMPOSER_AUTH` from the Terminus configuration yaml file at `~/.terminus/config.yml`
  - If the value is a string, assume a JSON string
  - If the value is not a string, parse the yaml into a JSON object
- Update documentation on how to use Composer auth

Out of scope:
- Prompting the user for a Composer auth token during `build:project:create` interactively

aaronbauman:
Opened https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/277 against this feature request.

I was able to test fetching `COMPOSER_AUTH` from environment, but not sure how the best way to read `~/.terminus/config.yml`

danielbachhuber:
> I was able to test fetching `COMPOSER_AUTH` from environment, but not sure how the best way to read `~/.terminus/config.yml`

What were you stuck on?

aaronbauman:
> > I was able to test fetching `COMPOSER_AUTH` from environment, but not sure how the best way to read `~/.terminus/config.yml`
> 
> What were you stuck on?

I ended up doing:
```
$this->session()->getConfig()->get('COMPOSER_AUTH');
```
Just wasn't sure if that's the recommended way to do it.

rvtraveller:
[...]

PR #277, opened 2019-09-17
Add support for COMPOSER_AUTH provider variable

aaronbauman:
Addresses https://github.com/pantheon-systems/terminus-build-tools-plugin/issues/275


danielbachhuber:
Thanks for your work on this, @aaronbauman.

I took what you did and expanded upon it in #292. Notably, the environment variable is `TERMINUS_BUILD_TOOLS_COMPOSER_AUTH` so that it's an explicit decision to export it to CircleCI, etc.

PR #279, opened 2019-09-23
Introduce `--use-ssh` to perform initial repo push over SSH

### LINKED ISSUE #276: Allow for ssh instead of https for the initial push to external Git providers during build:project:create
When running `build:project:create` users may prefer to make the initial commit to the Git provider by using SSH instead of `https`. In the case of self-hosted GitLab, this may be a forced requirement.

danielbachhuber:
* If `--use-ssh` is provided, command runs `$this->git_provider->verifySSHConnect()` to verify the process can connect to the provider over SSH.
* Default SSH URLs are: `git@bitbucket.org`, `git@github.com`, and `git@gitlab.com`. A self-hosted GitLab instance is assumed to be `git@selfhosteddomain.com`.
* `$this->git_provider->pushRepository()` has been adapted to simply `git push` for `--use-ssh`, because the Git remote is already being set on the repo.

Fixes #276


ataylorme:
@danielbachhuber I'm on the fence of whether we need test coverage variants for this. What do you think?

danielbachhuber:
> @danielbachhuber I'm on the fence of whether we need test coverage variants for this. What do you think?

What if we adapted the existing `test_wordpress_*` jobs to use `--ssh`, such that the Drupal jobs didn't and the WordPress jobs did?

ataylorme:
> What if we adapted the existing test_wordpress_* jobs to use --ssh, such that the Drupal jobs didn't and the WordPress jobs did?

That approach sounds great. It would give us coverage for both options over each repository provider.

danielbachhuber:
@ataylorme Looks like the CircleCI SSH key needs updated permissions in order to `--use-ssh`:

```
 20/24 [=======================>----]  83% [notice] Executing git push --progress origin master
ERROR: The key you are authenticating with has been marked as read only.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists
```

ataylorme:
```
Executing git push --progress origin master
ERROR: The key you are authenticating with has been marked as read only.
fatal: Could not read from remote repository.
```

@danielbachhuber it seems the SSH key CircleCI has set up is read-only. I'll have to coordinate with @greg-1-anderson to get that updated.

danielbachhuber:
@ataylorme Should I revert for now, and we can add it back later?

ataylorme:
> Should I revert for now, and we can add it back later?

No, I don't think we should merge the PR without test coverage, so that means fixing the tests

danielbachhuber:
@greg-1-anderson Any luck getting a new SSH key in place?

ataylorme:
@danielbachhuber access to GitLab and BitBucket is blocked right now. I'm trying to get it unblocked but the accounts Greg created were before we implemented Okta SSO and are now locked behind that.

In the meantime, since GitHub tests are passing, I say we:
- Revert the change to use SSH in GitLab and BitBucket tests
- Merge this PR
- Create a follow up issue to implement SSH testing for GitLab and BitBucket

danielbachhuber:
> Create a follow up issue to implement SSH testing for GitLab and BitBucket

Created with #287

danielbachhuber:
@ataylorme I think this is ready for merge whenever you want to give it another review.

PR #280, opened 2019-09-23
Update 'build:env:create' to comment on a PR instead of a commit

### LINKED ISSUE #238: multidev link on pull request, instead of a commit comment
A link to the multidev being sent as a pull/merge request comment from `terminus build:env:create` is missing.

`build:env:create` calls `build:comment:add:commit`, which calls `$this->git_provider->commentOnCommit`.

Instead of commenting on the commit the multidev link I think `build:env:create` should add the multidev link comemnt to the pull/merge request.

**edit** on my personal site I am running an [additional `create-multidev` job](https://github.com/ataylorme/ataylorme-wordpress/blob/master/.ci/deploy/pantheon/create-multidev) early in the workflow (in parallel with static tests) that uses `hub` to leave the multidev link on the PR. 

Having that in `build:env:create` with support for all Git providers would be a better solution.

danielbachhuber:
* Adds a new `build:comment:add:pr` command for creating comments on PRs.
* Define a new abstract method for commenting on a PR, implemented by each `GitProvider`.
* Updates `build:env:create` to comment on the PR instead of commit.

To test:
1. Create a new project (e.g. `terminus build:project:create wp daniel-test-gh-2 --git=github`).
2. Disable CI provider for the repository, because you'll want to run build tools locally.
3. Clone the new repository locally. Create a new branch, make some change to the code, and create a pull request from the branch.
4. Run `terminus build:env:create daniel-test-gh-2.dev pr-<num>` in the cloned repo, where `<num>` is the pull request number.

Fixes #238

image

image


src/Commands/EnvCreateCommand.php
```diff
@@ -150,18 +150,24 @@ public function createBuildEnv(
             $site_name = $site->getName();
             $project = $this->projectFromRemoteUrl($metadata['url']);
             $dashboard_url = "https://dashboard.pantheon.io/sites/{$site_id}#{$multidev}";
+            $pr_id = '';
+            if (preg_match('#^pr-([\d]+)$#', $multidev, $matches)) {
+                $pr_id = $matches[1];
+            }
```
danielbachhuber: @greg-1-anderson @ataylorme The one thing I don't love about this PR is pulling the `$pr_id` from `$multidev`.

The existing implementation is a little bit more robust because the commit `sha` is the same between Pantheon and the Git provider. We don't have same level of precision for the pull request ID, however.

Open to your suggestions on how to best derive this value. Another option would be to read the `PR_NUMBER` environment variable directly, but I feel like that's even more magical.

src/Commands/CommentAddPullRequestCommand.php
```diff
@@ -0,0 +1,60 @@
...
+
+        // Create a Git repository service provider appropriate to the URL
+        $this->inferGitProviderFromUrl($remoteUrlFromGit);
+
+        // Ensure that credentials for the Git provider are available
+        $this->providerManager()->validateCredentials();
+
+        // Compile message
+        if (!empty($options['site_url'])) {
+            $message = "[![Visit Site](https://raw.githubusercontent.com/pantheon-systems/ci-drops-8/0.1.0/data/img/visit-site-36.png)](".$options['site_url'].")\n\n".$options['message'];
```
ataylorme: We should move the `.png` image into the Build Tools repo, rather than relying on another repository.

src/Commands/CommentAddPullRequestCommand.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        // Create a Git repository service provider appropriate to the URL
+        $this->inferGitProviderFromUrl($remoteUrlFromGit);
+
+        // Ensure that credentials for the Git provider are available
+        $this->providerManager()->validateCredentials();
+
+        // Compile message
+        if (!empty($options['site_url'])) {
```
ataylorme: I think we should prioritize a user-passed message first, then fall back to constructing a message from `site_url`.

src/Commands/EnvCreateCommand.php
ataylorme: @danielbachhuber I think this is okay since `$doNotify` is only true when a new multidev is created. The multidev environment is `pr-` by default, though users can configure it to be the `ci-` variant.

Go ahead and fetch the PR number from the multidev name, as you are doing now, as well as from the environment variable and only commented on the pull request if they match.

danielbachhuber: > Go ahead and fetch the PR number from the multidev name, as you are doing now, as well as from the environment variable and only commented on the pull request if they match.

Which environment variable? It seems like there are at least a couple and I'm not sure which we need to respect.

src/Commands/CommentAddPullRequestCommand.php
danielbachhuber: I copied this from `CommentAddCommitCommand`. Should we change that as well for consistency, or should we make this change in a follow-up PR?

src/Commands/CommentAddPullRequestCommand.php
danielbachhuber: Do you have a preference as to where I put it? It doesn't look like we have any assets in this repository right now.

src/Commands/EnvCreateCommand.php
ataylorme: Either `PR_NUMBER` or `TERMINUS_ENV` would work but go with `TERMINUS_ENV` because it is required, where users may not have `PR_NUMBER` set. `TERMINUS_ENVV` will have the `pr-` prefix so you will need to parse the ID from that, just like you are doing with `$multidev`

src/Commands/CommentAddPullRequestCommand.php
ataylorme: `assets/images` works

src/Commands/CommentAddPullRequestCommand.php
ataylorme: A follow-up PR sounds good to reduce bloat here.

src/Commands/CommentAddPullRequestCommand.php
danielbachhuber: Updated in a65beefbcd1d20affdae8e8e1cd555e57b283a40 but also created #281

src/Commands/EnvCreateCommand.php
danielbachhuber: > `TERMINUS_ENV` because it is required,

Why is `TERMINUS_ENV` required? I didn't have it set locally when it was testing.

On second thought, I don't think we need to inspect an environment variable too. Unless @greg-1-anderson has some alternative idea, I'm happy with the current implementation.

ataylorme: >  I don't think we need to inspect an environment variable too

I'm okay with that.

danielbachhuber:
@greg-1-anderson Going to wait to merge until I get confirmation from you that there isn't a better way of accessing the PR number.

src/Commands/EnvCreateCommand.php
```diff
@@ -150,18 +150,24 @@ public function createBuildEnv(
...
                 'site' => $site_name,
                 'env' => $multidev,
+                'pr_id' => $pr_id,
                 'label' => $env_label,
                 'dashboard-url' => $dashboard_url,
                 'site-url' => "https://{$multidev}-{$site_name}.pantheonsite.io/",
                 'message' => "Created multidev environment [{$multidev}]({$dashboard_url}) for {$site_name}."
             ];
 
-            $command = $this->interpolate('terminus build:comment:add:commit --sha [[sha]] --message [[message]] --site_url [[site-url]]', $metadata);
```
greg-1-anderson: It seems to me that we should not eliminate the existing functionality of attaching a commit to an sha. It would be better if `build:commenet:add:commit` were used when the PR number is not known, and the new command used when the PR number is available.

src/Commands/EnvCreateCommand.php
```diff
@@ -150,18 +150,24 @@ public function createBuildEnv(
             $site_name = $site->getName();
             $project = $this->projectFromRemoteUrl($metadata['url']);
             $dashboard_url = "https://dashboard.pantheon.io/sites/{$site_id}#{$multidev}";
+            $pr_id = '';
+            if (preg_match('#^pr-([\d]+)$#', $multidev, $matches)) {
```
greg-1-anderson: This is certainly a possibility. Have you considered adding the pull request number as a new parameter to build:env:create? Do we already have a unified PR number available in the CI scripts in example-drops-8-composer & c.? If so, it would be pretty easy to update the CI scripts to pass this option in.

Another potential way to do this would be to ask the repository provider for the PR number, and have each repository provider examine the appropriately-named environment variable with this information. The environment variables set by the CI service should still be available to the Terminus Build Tools plugin commands. This would be quite a bit more code than the existing implementation (or the other proposed alternative), and I am unsure that there would be enough benefit of doing it this way to justify the cost.

src/Commands/EnvCreateCommand.php
danielbachhuber: > It would be better if `build:commenet:add:commit` were used when the PR number is not known, and the new command used when the PR number is available.

Is it expected behavior to magically switch between the two? I think it'd better to be an explicit decision.

greg-1-anderson: If your site is set up per the default configuration, then the pull request number will always be available. If a build runs before a PR is available, though, then what is the fallback behavior? I'd recommend thinking of comments on commits in that way, as fallback behavior, rather than as an alternative.

danielbachhuber: What about:

1. If `--pr-id=<num>` is specified as an argument, then we add a comment on the PR.
2. If no `--pr-id=<num>` is specified, then we add a comment on the changeset.

In this scenario, it'd be the responsibility of the calling script to indicate whether the comment is expected on the pull request vs. the changeset.

danielbachhuber: > In this scenario, it'd be the responsibility of the calling script to indicate whether the comment is expected on the pull request vs. the changeset.

... which is what I think you were hinting at in your comment below.

greg-1-anderson: Yes, I think that you should add the comment to the commit if the PR number is not available.

Whether you get the PR via the current method proposed in this PR, or one of the alternative ways that I suggested, depends on which is most reliable. If all of the three providers we're supporting today have environment variables with the PR number (I believe they all do), and if there are no barriers to passing the PR number in via an option, then I'd recommend you go that way. Otherwise, I'm okay with sticking with the regex-based PR number extraction. The multidev name might conceivably be renamed to something other than PR-nnn (if the user customized their scripts), so I'd guess that using the environment variables would probably be more reliable.

PR #282, opened 2019-09-24
Optionally use prepare for pantheon to skip force add during build:env:create and build:env:push

NickDickinsonWilde:
There are lots times that using `git add --force` is problematic. Using the prepare-for-pantheon architecture that has already been used in other projects is quite helpful. This implementing it as an option for env:create/env:push


greg-1-anderson:
Is this perhaps a duplicate of #202?

src/Commands/BuildToolsBase.php
```diff
@@ -813,7 +814,13 @@ public function pushCodeToPantheon(
         // any unwanted files prior to the build step (e.g. after a clean
         // checkout in a CI environment.)
         $this->passthru("git -C $repositoryDir checkout -B $branch");
-        $this->passthru("git -C $repositoryDir add --force -A .");
+        if ($prepareForPantheon) {
+          $this->passthru('composer prepare-for-pantheon');
```
ataylorme: This assumes that Composer has a `prepare-for-pantheon` script available

NickWilde1990:
yes @ataylorme that is the case. Since it is a non-default behaviour (and is noted that it runs that specific command in the option documention in `env:push` and `env:create`) is that okay or does it need some sort of error prevention/checking?

ataylorme:
@NickWilde1990 I don't think it would hurt to add error prevention. You can [reuse this logic from `build:project:create`](https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/master/src/Commands/ProjectCreateCommand.php#L365:L370) that checks for `composer build-assets` before running it.

src/Commands/BuildToolsBase.php
```diff
@@ -815,7 +815,11 @@ public function pushCodeToPantheon(
         // checkout in a CI environment.)
         $this->passthru("git -C $repositoryDir checkout -B $branch");
         if ($prepareForPantheon) {
-          $this->passthru('composer prepare-for-pantheon');
+          $this->log()->notice('Determine whether prepare-for-pantheon exists for {site}', ['site' => $site]);
+          exec("composer help build-assets", $outputLines, $status);
```
ataylorme: This should be `composer help prepare-for-pantheon`, not `composer help build-assets`

src/Commands/BuildToolsBase.php
```diff
@@ -815,7 +815,11 @@ public function pushCodeToPantheon(
         // checkout in a CI environment.)
         $this->passthru("git -C $repositoryDir checkout -B $branch");
         if ($prepareForPantheon) {
-          $this->passthru('composer prepare-for-pantheon');
+          $this->log()->notice('Determine whether prepare-for-pantheon exists for {site}', ['site' => $site]);
```
ataylorme: I don't think we need to log a notice at this point. Perhaps we can log a notice and fail or fall back to the current behavior if `preapre-for-pantheon` does not exist?

src/Commands/BuildToolsBase.php
NickWilde1990: Doh! thanks for catching that.

src/Commands/BuildToolsBase.php
NickWilde1990: pushed new changes

greg-1-anderson:
@NickWilde1990 Could you please be more specific about why you do not want to `force` during `git add`?  Are there files being produced by your build that you do not want committed to your repository?  Could you give an example?

NickWilde1990:
@greg-1-anderson Sure thing.
I'm doing some pattern lab stuff. So I have about 5-10k npm downloaded files, tempory files from the pattern lab build, as well as some sketchy gitignores that basically remove all the test files/folders from Drupal Core as well as vendor dependencies.
I could delete files in my CI process instead, but seemed easier to just use the gitignores.

PR #292, opened 2019-10-08
Support private Composer repos with `TERMINUS_BUILD_TOOLS_COMPOSER_AUTH` env variable

### LINKED ISSUE #275: Support for private packages

danielbachhuber:
Originally #277 
Fixes #275

Introduces a new `TERMINUS_BUILD_TOOLS_COMPOSER_AUTH` environment variable that will be passed through to `terminus build:project:create`. Alternatively, the value can be set in `~/.terminus/config.yml` with:

```yml
build-tools:
  composer-auth: '{ "http-basic": { "privatepackage.com": { "username": "daniel@handbuilt.co", "password": "abc123" } } }'
```


aaronbauman:
Not sure if it's a regression or this has never worked, but the env-based COMPOSER_AUTH option is not working for me on Bitbucket Pipelines.

I've set all the following environment variables, and none of them end up in my pipelines variables:
- `COMPOSER_AUTH` 
- `BUILD_TOOLS_COMPOSER_AUTH`
- `TERMINUS_BUILD_TOOLS_COMPOSER_AUTH`

PR #293, opened 2019-10-14
Add documentation for build:project:create options

aaronbauman:
Adds documentation for all options in `build:project:create` command.

### DIFF
--- a/src/Commands/ProjectCreateCommand.php
+++ b/src/Commands/ProjectCreateCommand.php
@@ -190,11 +190,22 @@ class ProjectCreateCommand extends BuildToolsBase
      * @param string $target Simple name of project to create.
      * @option org Organization for the new project (defaults to authenticated user)
      * @option team Pantheon team
+     * @option label The friendly name to use for the Pantheon site (defaults to the name of the Git site).
      * @option pantheon-site Name of Pantheon site to create (defaults to 'target' argument)
      * @option email email address to place in ssh-key
+     * @option test-site-name The name to use when installing the test site.
+     * @option admin-email The email address to use for the admin.
+     * @option admin-password The password to use for the admin when installing the test site.
      * @option stability Minimum allowed stability for template project.
+     * @option git Specify a git provider. Options are github (default), gitlab, and bitbucket.
+     * @option ci Specify a CI provider. Options are circleci, gitlab-pipelines, and bitbucket-pipelines. If not provided, CI will be assigned based on git provider choice.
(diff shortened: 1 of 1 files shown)


src/Commands/ProjectCreateCommand.php
```diff
@@ -190,11 +190,22 @@ public function validateCreateProject(CommandData $commandData)
      * @param string $target Simple name of project to create.
      * @option org Organization for the new project (defaults to authenticated user)
      * @option team Pantheon team
+     * @option label The friendly name to use for the Pantheon site (defaults to the name of the Git site).
      * @option pantheon-site Name of Pantheon site to create (defaults to 'target' argument)
      * @option email email address to place in ssh-key
+     * @option test-site-name The name to use when installing the test site.
+     * @option admin-email The email address to use for the admin.
+     * @option admin-password The password to use for the admin when installing the test site.
```
ataylorme: @aaronbauman a small nitpick here. Please add CMS so folks know this is the Drupal/WordPress admin password (I've seen people get confused and think these options are associated with their Pantheon account)

src/Commands/ProjectCreateCommand.php
```diff
@@ -190,11 +190,22 @@ public function validateCreateProject(CommandData $commandData)
...
      * @option stability Minimum allowed stability for template project.
+     * @option git Specify a git provider. Options are github (default), gitlab, and bitbucket.
+     * @option ci Specify a CI provider. Options are circleci, gitlab-pipelines, and bitbucket-pipelines. If not provided, CI will be assigned based on git provider choice.
      * @option visibility The desired visibility of the provider repository. Options are public, internal, and private.
      * @option use-ssh Use SSH instead of HTTPS to create the provider repository.
+     * @option region Specify a data residency region. Options are US (United States, default), AU (Australia), CA (Canada), and EU (European Union).
```
ataylorme: Rather than list the regions here please link to https://pantheon.io/docs/regions#available-regions

The doc will always be up-to-date so we don't have to worry about making a PR here when regions change in the future

ataylorme:
@aaronbauman thank you for the PR! I suggested a few small tweaks

src/Commands/ProjectCreateCommand.php
aaronbauman: Added "CMS"

src/Commands/ProjectCreateCommand.php
aaronbauman: Replaced list of options with URL to pantheon docs.

PR #299, opened 2019-10-22
Allow source_branch and sha for GitLab branch name in branchesForPullRequests output

### LINKED ISSUE #298: GitLab branchesForPullRequests returns SHA, not branch
[`branchesForPullRequests` in `GitLabProvider.php` line `198`](https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/7791475ecfa24aca55b434d174c2901385d916a7/src/ServiceProviders/RepositoryProviders/GitLab/GitLabProvider.php#L198) has `$branch_name = $item['sha'];`, which returns the merge request SHA.

`source_branch` should probably be used instead. See https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests

ataylorme:
Closes #298.

### DIFF
--- a/src/ServiceProviders/RepositoryProviders/Bitbucket/BitbucketProvider.php
+++ b/src/ServiceProviders/RepositoryProviders/Bitbucket/BitbucketProvider.php
@@ -131,7 +131,7 @@ class BitbucketProvider extends BaseGitProvider implements GitProvider, LoggerAw
     /**
      * @inheritdoc
      */
-    public function branchesForPullRequests($target_project, $state, $callback = null)
+    public function branchesForPullRequests($target_project, $state, $callback = null, $return_key = null)
     {
         $stateParameters = [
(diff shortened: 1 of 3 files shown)


ataylorme:
@rvtraveller no rush on this. It isn't blocking the `terminus-clu-plugin` since that has to implement its own `branchesForPullRequests` methods.

I updated things to accept a `$return_key` parameter, which defaults to `sha`. There is also a new private `filterBranchList` for GitLab that returns either the merge request branch `sha`, branch name or all branch data, based on that param.

This should be backward compatible and allow for more flexibility in the future.

However, if we abstract the providers out of Build Tools and into their own packages that Build Tools, `terminus-clu-plugin`, and others can consume in the future for a `3.0` it will be an issue.

ataylorme:
@danielbachhuber if you have time to do a code review on this before our meeting Friday that would be excellent.

PR #316, opened 2019-12-22
Project convert command

rvtraveller:
Initial implementation of a command to convert a project to the Git PR workflow. This command makes its best efforts to:

- Accept a source repository either in Pantheon or in a supported Git Provider
- Composerize the site
- Setup CI workflow in the user's chosen Git Provider (either what they used to start or what they specify via CLI if they are starting with a Pantheon site)
- Push back to the Git repository and let CI deploy it to Pantheon

I say best effort because there are plenty of different variations that could be input here and I'm sure there are plenty I haven't accounted for. I've assumed the most common use case is someone who started a Drupal site on Pantheon and now needs to switch off because they are trying to use Composer.


rvtraveller:
Through manual testing, this has an issue with some WP sites. There is a plan in place to address...I just need to execute on it based on the Version Tool work that is pending. I'm going to mark this as a WIP for now.

PR #325, opened 2020-02-19
Added support for changing the admin username

taeo920:
We tend to avoid using "admin" as a username so I've added support for providing a custom admin username via the build:project:create option with the key "admin-username".

I also cleaned up the getters and setters as the parameters were named incorrectly.

### DIFF
--- a/src/ServiceProviders/SiteProviders/PantheonProvider.php
+++ b/src/ServiceProviders/SiteProviders/PantheonProvider.php
@@ -98,7 +98,7 @@ class PantheonProvider implements SiteProvider, CredentialClientInterface, Publi
         $adminEmailRequest = (new CredentialRequest('ADMIN_EMAIL'))
             ->setRequired(false);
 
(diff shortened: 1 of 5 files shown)


src/ServiceProviders/SiteProviders/PantheonProvider.php
```diff
@@ -98,6 +98,9 @@ public function credentialRequests()
         $adminEmailRequest = (new CredentialRequest('ADMIN_EMAIL'))
             ->setRequired(false);
 
+        $adminEmailRequest = (new CredentialRequest('ADMIN_USERNAME'))
```
rvtraveller: Copy/paste issue? I'm assuming this should be `$adminUsernameRequest`

taeo920: Whoops yes this was a copy/paste error. I've corrected the issue and performed some manual test to ensure it didn't introduce any issues both with the username specified and without.

greg-1-anderson:
Thanks for the enhancement.

PR #330, opened 2020-03-10
Support WP in ProjectCreateCommand install

NickDickinsonWilde:
`build:project:create` command doesn't work for WP - attempt to use Drush.
And had a rather suspicious line: 
`// TODO: Select a different default template based on the cms type (Drupal or WordPress).`

PR implements that using the existing detection method.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -559,9 +559,10 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
             'account-pass' => '',
             'site-mail' => '',
             'site-name' => ''
-        ])
+        ],
+        $upstream = 'drupal')
     {
-        $command_template = $this->getInstallCommandTemplate($composer_json);
+        $command_template = $this->getInstallCommandTemplate($composer_json, $upstream);
         return $this->runCommandTemplateOnRemoteEnv($site_env_id, $command_template, "Install site", $site_install_options);
     }
 
@@ -670,15 +671,15 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
     /**
      * Determine the command to use to install the site.
      */
-    protected function getInstallCommandTemplate($composer_json)
+    protected function getInstallCommandTemplate($composer_json, $upstream)
     {
         if (isset($composer_json['extra']['build-env']['install-cms'])) {
             return $composer_json['extra']['build-env']['install-cms'];
         }
-        // TODO: Select a different default template based on the cms type (Drupal or WordPress).
-        $defaultTemplate = 'drush site-install --yes --account-mail={account-mail} --account-name={account-name} --account-pass={account-pass} --site-mail={site-mail} --site-name={site-name}';
-
-        return $defaultTemplate;
+        if ($upstream == 'empty-wordpress') {
+            return 'wp core install --url={site-url} --title="{site-name}" --admin_user={account-name} --admin_password={account-pass} --admin_email={account-mail}';
+        }
+        return 'drush site-install --yes --account-mail={account-mail} --account-name={account-name} --account-pass={account-pass} --site-mail={site-mail} --site-name={site-name}';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
Build tools expects this in your [WordPress project template](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/composer.json#L53).

I'm not opposed to having a default for WordPress in addition to Drush, but I'm not thrilled about having a separate upstream option. Maybe we could use the [version tool](https://github.com/consolidation/version-tool) to figure out what the framework is, if we need to pick a default?

cc @sugaroverflow

greg-1-anderson:
Oh, I misread the patch. This is just adding another parameter to an internal function; it's not a new flag to the command. That's not as bad.

src/Commands/BuildToolsBase.php
```diff
@@ -670,15 +671,15 @@ protected function redactMetadata($metadata, $keys_to_redact)
...
+    protected function getInstallCommandTemplate($composer_json, $upstream)
     {
         if (isset($composer_json['extra']['build-env']['install-cms'])) {
             return $composer_json['extra']['build-env']['install-cms'];
         }
-        // TODO: Select a different default template based on the cms type (Drupal or WordPress).
-        $defaultTemplate = 'drush site-install --yes --account-mail={account-mail} --account-name={account-name} --account-pass={account-pass} --site-mail={site-mail} --site-name={site-name}';
-
-        return $defaultTemplate;
+        if ($upstream == 'empty-wordpress') {
```
rvtraveller: We'll need to find a different way to detect WP or Drupal. There are plenty of people out there that use Build Tools with WP upstreams that aren't `empty-wordpress`.

rvtraveller: Actually, this might not be accurate. One moment...

rvtraveller: Yup, nevermind here. `autoDetectUpstream()` returns a fixed list.

greg-1-anderson: version-tool is a more accurate way to find the framework type.

nickwilde1990: > Build tools expects this in your [WordPress project template](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/composer.json#L53)
>
Yep, I saw that through reading the function - not documented anywhere I can see - but figured a default WP and a default Drupal option covers the basics better, especially with the comment there.

re: using version-tool or something. That fixed list from autoDetectUpstream does seem a bit weak, but figured as it was being used elsewhere that was the most minimal change to be fully functional.

re: next steps, would you suggest switching to the composer.json method and not setting a default? (if so, want a documentation PR about that?). Or do you think having a default is a good direction?

nickwilde1990:
> Build tools expects this in your [WordPress project template](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/composer.json#L53)

Yep, I saw that through reading the function - not documented anywhere I can see - but figured a default WP and a default Drupal option covers the basics better, especially with the comment there.

re: using version-tool or something. That fixed list from autoDetectUpstream does seem a bit weak, but figured as it was being used elsewhere that was the most minimal change to be fully functional.

re: next steps, would you suggest switching to the composer.json method and not setting a default? (if so, want a documentation PR about that?). Or do you think having a default is a good direction?

greg-1-anderson:
re: next steps, I will defer to @sugaroverflow on that

sugaroverflow:
👋 hello friends, I'm the new maintainer for build tools and still getting up to speed on the plugin and how things work. I don't completely understand what's being discussed here so I will need a bit more context :)

Questions:

> Build tools expects this in your [WordPress project template.](https://github.com/pantheon-systems/example-wordpress-composer/blob/master/composer.json#L53)

@greg-1-anderson  does `"install-cms":` take the place of `build:project:create`? so all wordpress sites are expected to use their composer.json to be installed while drupal sites are installed by drush in build tools?

Also, what is the version-tool being referenced?

nickwilde1990:
Congrats on the maintainership Fatima!
Greg may have more but:
version-tool: https://github.com/consolidation/version-tool - an standalone of Drupal (6-8) or WP detector

greg-1-anderson:
@sugaroverflow What @nickwilde1990 said is correct. We can go over this at our next sync meeting.

sugaroverflow:
note - this is waiting on #331 and will likely get in for our 2.0 :)

kporras07:
This is currently being worked on https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/401 

Once merged, we could close this issue.

stevector:
https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/401 was merged.

PR #331, opened 2020-03-17
Introduce Version Tool into Build Tools.

rvtraveller:
Let's introduce the [Version Tool](https://github.com/consolidation/version-tool) package into Build Tools so we can leverage it in other PRs or work we're doing.

This includes a `build:project:version` command which is largely available just for testing purposes.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -394,7 +394,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $info = $this->autodetectApplication($siteDir);
 
         if ($info) {
-            return $info;
+            return $info['framework'];
         }
 
         // For backwards compatibility, assume "Empty".
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -30,6 +30,7 @@ use Pantheon\Terminus\Models\Environment;
 
 use Robo\Contract\BuilderAwareInterface;
 use Robo\LoadAllTasks;
+use VersionTool\VersionTool;
 
 /**
  * Build Tool Base Command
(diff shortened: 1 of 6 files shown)


src/Commands/ProjectVersionCommand.php
```diff
@@ -0,0 +1,42 @@
...
+    use \Pantheon\TerminusBuildTools\Task\Ssh\Tasks;
+    use \Pantheon\TerminusBuildTools\Task\CI\Tasks;
+
+    /**
+     * Show build info attached to a site created by the
+     * build:project:create command.
+     *
+     * @command build:project:version
+     */
+    public function version()
```
greg-1-anderson, line 26: Please see https://github.com/consolidation/version-tool/blob/master/src/Cli/InfoCommands.php

greg-1-anderson: It would be best to return structured data so that --fields et. al. are available

rvtraveller: Thanks. I updated accordingly.

sugaroverflow:
@rvtraveller made some changes - @greg-1-anderson could you review if this is good to be merged now? Thank you!

src/Commands/BuildToolsBase.php
```diff
@@ -343,41 +344,64 @@ protected function expandSourceAliases($source)
...
 
-        foreach ($upstream_map as $file => $upstream) {
-            if (file_exists("$siteDir/$file")) {
-                return $upstream;
-            }
+        if ($info) {
+            $application = $info->application();
+            $version = $info->version();
+            $major_version = explode(".", $version);
+            $framework = $application == "WordPress" ? "empty-wordpress" : (
```
greg-1-anderson, line 374: This could be a little more verbose for readability. Maybe break it out into its own function where you can use the pattern `if (cond) { return }` to avoid nesting and ternary operators.

src/Commands/ProjectVersionCommand.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $info = $this->autodetectApplication(getcwd());
+
+        if ($info) {
+            $data = [
+              'application' => $info['application'],
+              'version' => $info['version'],
+              'major-version' => $info['major_version'],
```
greg-1-anderson, line 36: If we used `$info['major-version']` instead of `major_version` throughout the code, then we could just return $info here without having to create a mapping like this.

PR #335, opened 2020-03-30
Only override GIT_SSH_COMMAND if the user has not already set it

### LINKED ISSUE #334: Allow overrides to GIT_SSH_COMMAND
Hey Folks!

I am using the build tools plugin inside my own Drupal CI service that I am building to easily support deploys to Pantheon. One of the things my CI service does is override `$GIT_SSH_COMMAND` to inject deploy keys when pulling from Github and pushing to Pantheon. This works fine on the clone from Github side, but it fails when I go to push to Pantheon because of https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/52cc3ee9ed0c3e9765d3c29a67fa07bea2c2e28a/src/Commands/BuildToolsBase.php#L76

In my own code I am doing this: 
```php
                          [
                                    'name' => 'GIT_SSH_COMMAND',
                                    'value' => implode(" ", [
                                        'ssh',
                                        '-i',
                                        '/deploy_key/private_key',
                                        '-o',
                                        'UserKnownHostsFile=/dev/null',
                                        '-o StrictHostKeyChecking=no'
                                    ])
                                ],
```
So when Terminus throws the putenv in here, it causes sadness and gnashing of teeth.

Proposed solution:
Check if the environment variable is already defined, and only inject it if it is empty. Alternatively, I could see providing a flag/option to the commands that would allow for overriding of the SSH command.

dustinleblanc:
This could resolve #334

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -80,7 +80,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         // Setting the Known Hosts File to /dev/null and the LogLevel to quiet prevents
         // this from persisting for a user regularly as well as the warning about adding
         // the SSH key to the known hosts file.
-        if (getenv('GIT_SSH_COMMAND') !== false) {
+        if (getenv('GIT_SSH_COMMAND') === false) {
             putenv("GIT_SSH_COMMAND=ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=QUIET");
         }
     }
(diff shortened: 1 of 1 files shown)


rvtraveller:
I like this approach and am generally on board with the change. The one thing that I think would be useful is to print something out when the user has the `GIT_SSH_COMMAND` environment variable set. Something like "Using your custom GIT_SSH_COMMAND variable instead of the Build Tools default" or similar.

greg-1-anderson:
I'd recommend that we instead get the user's GIT_SSH_COMMAND setting and then iterate through each of our custom entries:

- `-o UserKnownHostsFile=/dev/null` 
- `-o StrictHostKeyChecking=no`
- `-o LogLevel=QUIET`

We could make a map with entries like `'UserKnownHostsFile' => '/dev/null'`, and then composer the `-o KEY=VALUE` line if KEY does not already appear in the user's GIT_SSH_COMMAND.

That would be safer, and more likely to produce the desired result.

If GIT_SSH_COMMAND is not defined, we could just start with `ssh` and then run the same algorithm to add all of the config items.

Issue #281, opened 2019-09-23
Update 'visit-site-36.png' path to refer to tagged version


danielbachhuber:
Immediately prior to releasing v2.0.0, we should update 'visit-site-36.png' references (a65beef) to use the tagged URL. This way, if we ever change the path on the master branch, it won't break existing usage.

PR #343, opened 2020-05-01
Replace the Visit Site image path

sugaroverflow:
Solves #281 to update the Visit Site path with the tagged release path instead of master in case we change it in the future.


rvtraveller (changes requested):
I think we need to include `?raw=true` at the end of our URLs so we get the raw image.

sugaroverflow:
Gotcha, changes made.

greg-1-anderson (approved):

rvtraveller (approved):
LGTM!

PR #346, opened 2020-06-08
Adding an option for --no-git-force on build:env:create and build:env:push

stevector:
This PR is a replacement for https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/202 which has been used by our CircleCI Orb for the last year and is far behind master. The motivation here is to be able to use the Build Tools with little, or no risk of a force push overwriting history.

To test this change, I [also made a PR on the orb to use this version of Build Tools](https://github.com/pantheon-systems/circleci-orb/pull/45).
I then used that version of the orb in [a build](https://app.circleci.com/pipelines/github/stevector/nerdologues-d8/347/workflows/ceb0a006-0568-40c7-a612-6afbe0f7719c/jobs/4271) on a [consumer of the Orb](https://github.com/stevector/nerdologues-d8/pull/366/commits/8516b20965dfa445cdabcefa331cf4aa7efc523c).

Here's a screenshot of the section of the build where add and push happened without `--force`
Screen Shot 2020-06-08 at 8 58 26 PM

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -728,7 +728,8 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $multidev = '',
         $repositoryDir = '',
         $label = '',
-        $message = '')
+        $message = '',
+        $gitForce = FALSE)
     {
         list($site, $env) = $this->getSiteEnv($site_env_id);
         $dev_env = $site->getEnvironments()->get('dev');
@@ -807,7 +808,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         // any unwanted files prior to the build step (e.g. after a clean
         // checkout in a CI environment.)
         $this->passthru("git -C $repositoryDir checkout -B $branch");
-        if ($this->respectGitignore($repositoryDir)) {
+        if ($this->respectGitignore($repositoryDir) || $gitForce === FALSE) {
             // In "Integrated Composer" mode, we will not commit ignored files
             $this->passthru("git -C $repositoryDir add .");
         }
@@ -827,7 +828,8 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
 
         // Push the branch to Pantheon
         $preCommitTime = time();
-        $this->passthru("git -C $repositoryDir push --force -q pantheon $branch");
+        $forceFlag = $gitForce ? "--force" : "";
+        $this->passthru("git -C $repositoryDir push $forceFlag -q pantheon $branch");
 
         // If the environment already existed, then we risk encountering
         // a race condition, because the 'git push' above will fire off
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -729,7 +729,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $repositoryDir = '',
         $label = '',
         $message = '',
-        $gitForce = FALSE)
+        $gitForce = TRUE)
     {
         list($site, $env) = $this->getSiteEnv($site_env_id);
         $dev_env = $site->getEnvironments()->get('dev');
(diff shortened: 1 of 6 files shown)


stevector:
The Drupal 8 tests are failing because `example-drops-8-composer` now requires `"php": ">=7.3"` but the container here uses `7.2`. @greg-1-anderson is there a container upgrade to make?

stevector:
ah, we need to change this: https://github.com/pantheon-systems/docker-php-ci/blob/2.x/Dockerfile#L2

src/Commands/EnvPushCommand.php
```diff
@@ -22,15 +22,19 @@ class EnvPushCommand extends BuildToolsBase
      *
      * @param string $site_env_id Site and environment to push to. May be any dev or multidev environment.
      * @param string $repositoryDir Code to push. Defaults to cwd.
+     * @option label What to name the environment in commit comments
```
stevector, line 25: These two lines for `label` and `message` are not material to why I made this PR. But it seemed like a bug to be omitting them as command line options when they are present as options in the function signature.

stevector:
@greg-1-anderson and @sugaroverflow, I added this PR to the agenda for our Wednesday meeting.

src/Commands/BuildToolsBase.php
```diff
@@ -728,7 +728,8 @@ public function pushCodeToPantheon(
         $multidev = '',
         $repositoryDir = '',
         $label = '',
-        $message = '')
+        $message = '',
+        $NoGitForce = FALSE)
```
greg-1-anderson, line 732: Use camelCase

greg-1-anderson:
I'm +1 on this in concept. Did not look at the test failures; those should be fixed or at least explained. Could be transient.

sugaroverflow:
+1 this makes sense to me! Can we also update the readme with the new command line option?

src/Commands/BuildToolsBase.php
stevector: Good catch, pushing a fix.

greg-1-anderson:
@stevector It took some fiddling, since our upstream image switched to Alpine, but I made a 4.x branch working with PHP 7.3.  Give it a try here and see if it clears up the test failures.

stevector:
@sugaroverflow, I updated the readme.

@greg-1-anderson, I also switched to the 4.x container and that broke the set up scripts but I'm not sure why beyond the switch to Alpine.

greg-1-anderson:
Probably missing some binaries we depend on...

sugaroverflow:
Yes, it looks like we're missing these two : 

```
/bin/sh: ./.circleci/cleanup-fixtures.sh: not found
/bin/sh: ./.circleci/set-up-globals.sh: not found
```

do they just need to be included in the circleCI folder?

greg-1-anderson:
Those scripts exist; the "not found" error probably comes from the shell not being able to find the bash interpreter listed on the shebang (`#!`) line.

stevector:
@sugaroverflow I re-triggered the failed build and we're all green here now!

sugaroverflow:
yay @stevector - I'll merge this today and get it in for this week's release :) thank you for your patience! 🚀

src/Commands/EnvCreateCommand.php
```diff
@@ -55,7 +57,7 @@ public function createBuildEnv(
         if ('dev' === $multidev) {
             $this->log()->notice('dev has been passed to the multidev option. Reverting to dev:env:push as dev is not a multidev environment.');
             // Run build:env:push.
-            $this->pushCodeToPantheon($site_env_id, $multidev, '', $env_label);
+            $this->pushCodeToPantheon($site_env_id, $multidev, '', $env_label, $options['message'], $options['git-force']);
```
stevector, line 60: @greg-1-anderson @sugaroverflow,it looks like this was a bug. This addition uses `git-force` when the param is actually `no-git-force`

greg-1-anderson: The right way to do this with the annotated command library is to use `'git-force' => true`. That will make `$options['git-force']` true by default, and will add a `--no-git-force` option to set it to false.

stevector: ok, so #361 needs more work to accommodate that then?

greg-1-anderson: You're right, I didn't notice the double-negative of $noGitForce in pushCodeToPantheon(). Added a `!` to the PR.

PR #348, opened 2020-06-23
Update composer install step to stable release

ccharlton:
I noticed the install snippet was still referencing the last pre-release (beta-17) instead of the 2.0.0 stable release.


greg-1-anderson:
LGTM, thanks.

PR #353, opened 2020-08-10
BUGS-3249: Increase workflow timeout and add better logging

jburianek:
Pantheon workflows can take longer than 60 seconds to complete, so let's bump up the timeout.

Also, we should add better logging around this.

Success after this change:
```
 [notice] Workflow 'Sync code on "dev"' running.
 [notice] Workflow 'Sync code on "dev"' succeeded.
 [notice] Workflow succeeded
```

Timeout after this change:
```
 [notice] Workflow 'Sync code on "dev"' running.
 [warning] Waited '2' seconds, giving up waiting for workflow to finish
```

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -997,14 +997,14 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
         $this->waitForWorkflow($startTime, $site, $env_name);
     }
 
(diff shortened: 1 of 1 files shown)


rvtraveller:
This looks fine to me (pending tests of course).

@sugaroverflow do you know if there are internal Pantheon metrics for how long this workflow takes on 99% and 95% time frames? Wondering what the normal time is and if additional messaging is necessary around how much time this normally takes?

PR #361, opened 2020-10-22
Fixes #358: git force flag

### LINKED ISSUE #358: Undefined index: git-force
This line https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/main/src/Commands/EnvCreateCommand.php#L60 causes _createBuildEnv_ to throw a notice like :
```
[notice] dev has been passed to the multidev option. Reverting to dev:env:push as dev is not a multidev environment.

Notice: Undefined index: git-force in /usr/local/share/terminus-plugins/terminus-build-tools-plugin/src/Commands/EnvCreateCommand.php on line 60

Call Stack:
    0.0003     427792   1. {main}() /usr/local/share/terminus/vendor/pantheon-systems/terminus/bin/terminus:0
    0.1099    6685888   2. Pantheon\Terminus\Terminus->run() /usr/local/share/terminus/vendor/pantheon-systems/terminus/bin/terminus:49
    0.1099    6685912   3. Robo\Runner->run() /usr/local/share/terminus/vendor/pantheon-systems/terminus/src/Terminus.php:108
    0.6021   10918448   4. Symfony\Component\Console\Application->run() /usr/local/share/terminus/vendor/consolidation/robo/src/Runner.php:234
    0.6087   10936528   5. Symfony\Component\Console\Application->doRun() /usr/local/share/terminus/vendor/symfony/console/Application.php:148
    0.6104   10939648   6. Symfony\Component\Console\Application->doRunCommand() /usr/local/share/terminus/vendor/symfony/console/Application.php:255
    0.6145   10980392   7. Consolidation\AnnotatedCommand\AnnotatedCommand->run() /usr/local/share/terminus/vendor/symfony/console/Application.php:1005
    0.6198   11091728   8. Consolidation\AnnotatedCommand\AnnotatedCommand->execute() /usr/local/share/terminus/vendor/symfony/console/Command/Command.php:255
    0.6203   11108344   9. Consolidation\AnnotatedCommand\CommandProcessor->process() /usr/local/share/terminus/vendor/consolidation/annotated-command/src/AnnotatedCommand.php:302
    0.6203   11108344  10. Consolidation\AnnotatedCommand\CommandProcessor->validateRunAndAlter() /usr/local/share/terminus/vendor/consolidation/annotated-command/src/CommandProcessor.php:176
    0.6212   11124704  11. Consolidation\AnnotatedCommand\CommandProcessor->runCommandCallback() /usr/local/share/terminus/vendor/consolidation/annotated-command/src/CommandProcessor.php:212
    0.6213   11126416  12. call_user_func_array:{/usr/local/share/terminus/vendor/consolidation/annotated-command/src/CommandProcessor.php:257}() /usr/local/share/terminus/vendor/consolidation/annotated-command/src/CommandProcessor.php:257
    0.6213   11126464  13. Pantheon\TerminusBuildTools\Commands\EnvCreateCommand->createBuildEnv() /usr/local/share/terminus/vendor/consolidation/annotated-command/src/CommandProcessor.php:257
```

Using it from the image _quay.io/pantheon-public/build-tools-ci:6.x_ on a CI/CD project while calling `terminus -n build:env:create ...`

greg-1-anderson:


stevector:
I'm not sure that this is the appropriate change. See https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/346/files#r510274896

greg-1-anderson:
Ah, I see, the option is reversed in `pushCodeToPantheon()`. In the short term it's probably best to just add a `!` at the point of call, although the double-negative is a bit confusing and might warrant cleaning up.

src/Commands/EnvCreateCommand.php
```diff
@@ -56,8 +56,8 @@ public function createBuildEnv(
         // Revert to build:env:push if build:env:create is run against dev.
         if ('dev' === $multidev) {
             $this->log()->notice('dev has been passed to the multidev option. Reverting to dev:env:push as dev is not a multidev environment.');
-            // Run build:env:push.
-            $this->pushCodeToPantheon($site_env_id, $multidev, '', $env_label, $options['message'], $options['git-force']);
+            // Run build:env:push. Note the double-negative on no-git-force.
```
stevector, line 59: `--no-git-force` is already in the docs. https://github.com/pantheon-systems/terminus-build-tools-plugin#command-options-5

I agree that the double negative is confusing. But at this point I think the easier fix is to change only this line to `no-git-force`

greg-1-anderson: As I said on the other issue, declaring `'git-force' => true` in $options creates a `--no-git-force` option that will set `$options['git-force']` to false if used.

The double-negative I was referring to was the variable used in pushCodeToPantheon.

If you define the option `'no-git-force'` instead of `'git-force'`, then the `--git-force` flag does not work. With the idiomatic solution, then you have both `--no-git-force` and `--git-force`.

So, in short, I still recommend the PR as currently written as the fix for this bug -- presuming that it works, of course.

sugaroverflow: I'd go with the simpler change if that resolves the bug.

greg-1-anderson: I don't have time to iterate on this right now, but I have no objection if someone wants to take this in a different direction

sugaroverflow: No worries, I'll give it a shot :)

sugaroverflow: After looking into the annotated command library (thanks Greg!), I'm not sure that I like the double negative solution. It feels confusing. Do we really need a `--git-force` flag?

greg-1-anderson: Yes. Terminus Build Tools must use force so that it can overwrite the last commit with the build artifacts. Therefore, force must be the default. @stevector also has a use-case where he would rather not --force. Also, when we build the integration between build tools and integrated composer, we will need to support the no-force option.

I suppose our other option is to do a major version bump and say that the next version is IC only, and never force there. That's an easy solution, but I think we're going to want to maintain traditional build tools in parallel with the IC-supporting version, and it doesn't seem worth maintaining two separate versions just for the force flag.

I would still recommend that we solve this with the PR as I wrote it. Note that there is no double-negative in the use of the annotated flag. --git-force is true, and --no-git-force is false, as you would expect. The double-negative is only in the pushCodeToPantheon method. We could easily get rid of that by inverting the name and meaning and every usage of the variable name, and remove the ! in the call to the method as well. I held off on doing that due to lack of time and to keep the diff minimal so that we could just commit here and get the fix in.

It's up to you how to continue, but I don't think that the quick fix is bad; we can clean up the additional tech debt later without adding to it if we commit this now.

stevector:
@sugaroverflow is this issue still on your radar?

sugaroverflow:
@stevector I've been struggling to find time to work on this. Because of this, I could merge this as is and open an issue to address the double negative in the future. Do you have any thoughts on that?

greg-1-anderson:
This PR is using `'git-force' =>  true,`, so I am +1 on merging it as-is. That's the important part of the change. I don't think it's too bad to leave the param names / meaning as is.

PR #362, opened 2020-10-27
BUGS-3249 Add TERMINUS_BUILD_TOOLS_WORKFLOW_TIMEOUT environment variable

marfillaster:


greg-1-anderson:
+1 on this; maybe we should also bump the default up to 300 or so.

sugaroverflow (approved):
Looks good to me thank you! We got a question about this today! Running tests in #364 and also bumped the default workflow timeout to 300. Merging shortly.

Issue #374, opened 2021-04-09
GitHub Personal Access Token Rejected


aweingarten:
When trying to use `terminus build:project:create --git=github` build tools no longer accepts github access tokens. It returns the following errror: `GitHub authentication tokens should be 40-character strings containing only the letters a-f and digits (0-9). Please enter your token again.`

It looks like GitHub changed their token format:
https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/

We will need to update the token validation code to handle the new format.

aweingarten:
Looks like the issue is that the regex here needs a simple-update:
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/0ae99c33fcae5b393947636f0a80e2049d1e51cd/src/API/GitHub/GitHubAPITrait.php#L75

sugaroverflow:
This got merged into `main` - I can cut a release later today (waiting in case we get that D9 support PR)

PR #375, opened 2021-04-09
ISSUE-374: Updated the GitHub regex to support their new personal access token format

aweingarten:
Addresses the issue documented here:
https://github.com/pantheon-systems/terminus-build-tools-plugin/issues/374


greg-1-anderson:
Thanks

PR #387, opened 2021-04-30
Fix branch name method to avoid infinite recursion

azeemhassni:
Fixes the `branchName` method in PullRequestInfo class to avoid infinite recursion.


greg-1-anderson:
I guess this was mostly unused? Thanks for catching this.

PR #390, opened 2021-07-15
[CMS-145] [CMS-146] Add template repository option to project create command.

kporras07:
# Private composer repository.

To use a private repository, define the auth string like this:

```
export TERMINUS_BUILD_TOOLS_COMPOSER_AUTH=json_encoded_string
```

or in .terminus/config.yml as described at https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/main/src/Utility/Config.php#L21

Then, in the build:project:create command, pass a composer-repository option like this:

```
terminus build:project:create --template-repository="https://repo.packagist.com/kporras07" kporras07/example-drops-8-test my-project
```

# Git repository (no composer repository)

To use a git repository, use the terminus command like this:

```
terminus build:project:create --template-repository="git@github.com:kporras07/myrepo.git" kporras07/myrepo-template my-project
```

The package name in the composer.json file into the template repo should be "kporras07/myrepo-template". If kporras07/myrepo is a private repo, you should have access to it in your current terminal.

You can also use it as the following shorthand:

```
terminus build:project:create git@github.com:kporras07/example-drops-8-test.git  buildtoolsd807148
```

During the project creation, the repo will be cloned to gather the right package name from the composer.json file.

### DIFF
--- a/src/Commands/BuildToolsBase.php
+++ b/src/Commands/BuildToolsBase.php
@@ -388,7 +388,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
             return $this->useExistingSourceDirectory($source, $options['preserve-local-repository']);
         }
         else {
-            return $this->createFromSourceProject($source, $target, $stability, $options['composer-repository']);
+            return $this->createFromSourceProject($source, $target, $stability, $options['composer-repository'], $options['template-git-url']);
         }
     }
 
@@ -410,7 +410,7 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
     /**
      * Use composer create-project to create a new local copy of the source project.
      */
-    protected function createFromSourceProject($source, $target, $stability = '', $composer_repository = '')
+    protected function createFromSourceProject($source, $target, $stability = '', $composer_repository = '', $template_git_url = '')
     {
         $source_project = $this->sourceProjectFromSource($source);
 
@@ -431,6 +431,10 @@ class BuildToolsBase extends TerminusCommand implements SiteAwareInterface, Buil
 
         if ($composer_repository) {
             $repository = ' --repository=' . $composer_repository;
+        }
+
+        if ($template_git_url) {
+            $repository = ' --repository="{\"url\": \"' . $template_git_url . '\", \"type\": \"vcs\"}"';
         }
 
         $this->passthru("composer create-project --working-dir=$tmpsitedir $repository $source $target -n $stability_flag");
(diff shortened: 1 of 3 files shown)


greg-1-anderson (changes requested):
This looks great. Could you also update the README with info from the instructions you included in this PR's conversation, and also a link to external documentation on setting up a packagist repo?

greg-1-anderson:
One more request: let's give folks a hint on where to find info on the authentication information. Maybe the best docs link would be: https://getcomposer.org/doc/articles/authentication-for-private-packages.md#authentication-in-auth-json-per-project

Feel free to substitute something else if there's a different location you think is more relevant.

kporras07:
@greg-1-anderson thanks. Done.

greg-1-anderson (approved):
Looks good to merge once tests pass.

kporras07:
@greg-1-anderson I updated the PR to include CMS-146

greg-1-anderson:
I was wondering if we could save the user from the need to specify the project name, i.e. instead of:
```
terminus build:project:create --template-repository="git@github.com:kporras07/myrepo.git" kporras07/myrepo-template my-project
```
How about:
```
terminus build:project:create git@github.com:kporras07/myrepo.git my-project
```
If build:project:create detects that the template parameter is a git URL instead of org/project, then use it to set the template repository and read the composer.json to get the project name. This will involve some redundant project access (once by us and once by Composer), but a clone with depth 0 shouldn't take very long.

This could perhaps be in addition to the current implementation, so that folks have the option of typing everything out if they want to avoid the extra download (on a slow connection or whatever).

What do you think about that?

kporras07:
Sounds good. I'll look into this.

And what if it's a private composer repository?  How should the command look?

greg-1-anderson:
The private composer repository will need both parameters, i.e. unchanged from the current implementation in this PR.

kporras07:
@greg-1-anderson updated. Please take another look.

src/Commands/BuildToolsBase.php
```diff
@@ -443,6 +462,39 @@ protected function sourceProjectFromSource($source)
...
+        ];
+        if (preg_match('/^[A-Za-z0-9\-]*\/[A-Za-z0-9\-]*:?[A-Za-z0-9\-]*$/', $source)) {
+            $items['source'] = $source;
+        }
+        else {
+            $items['template-repository'] = '{\"url\": \"' . $source . '\", \"type\": \"vcs\"}';
+            $templateDir = $this->tempdir('template-dir');
+            $this->passthru("git -C $templateDir clone $source --depth 1 .");
+            $composer_json_contents = file_get_contents($templateDir . '/composer.json');
+            if (preg_match('/"name":\s"([A-Za-z0-9\-]+\/[A-Za-z0-9\-]+)"/', $composer_json_contents, $matches)) {
```
greg-1-anderson, line 489: Please use json_decode here instead of a regex match.

greg-1-anderson:
@kporras07 I invited you to this repo; could you please re-submit this PR from a branch in this repository so that the tests can run. With the json_decode change and passing tests, I think this will be ready to go.

kporras07:
Closing in favor of #391

PR #392, opened 2021-07-27
Add ci from other repo

kporras07:
This PR allows you to create sites using templates that don't have CI in them (e.g. pantheon-upstreams/drupal-project). By default, it uses pantheon-systems/tbt-ci-integrations but it could be overriden using --ci-template option.

Example command:
```
terminus build:project:create git@github.com:pantheon-upstreams/drupal-project.git <project_name> --stability=dev
```

### DIFF
--- a/src/Commands/ProjectCreateCommand.php
+++ b/src/Commands/ProjectCreateCommand.php
@@ -143,14 +143,14 @@ class ProjectCreateCommand extends BuildToolsBase
     /**
      * Copy CI files from the given/default repo.
      */
-    public function copyCiFiles($ci_provider, $created_folder, $cms_version = 'd8', $repo = '') {
+    public function copyCiFiles($ci_provider, $created_folder, $cms_version, $ci_template = '') {
         $fs = new Filesystem();
         $service_name = $ci_provider->getServiceName();
-        if (!$repo) {
-            $repo = 'git@github.com:kporras07/tbt-ci-integrations.git';
+        if (!$ci_template) {
+            $ci_template = 'git@github.com:kporras07/tbt-ci-integrations.git';
         }
         $ciTemplateDir = $this->tempdir('ci-template-dir');
-        $this->passthru("git -C $ciTemplateDir clone $repo --depth 1 .");
+        $this->passthru("git -C $ciTemplateDir clone $ci_template --depth 1 .");
 
         $fs->mirror("$ciTemplateDir/$cms_version/.ci", "$created_folder/.ci");
         $fs->mirror("$ciTemplateDir/$cms_version/providers/$service_name/.", $created_folder);
@@ -237,10 +237,9 @@ class ProjectCreateCommand extends BuildToolsBase
      * @option keep If given, clone a local copy of the project.
      * @option env Add extra environment variables to the CI environment. For example, --env='key=value' --env='another=v2'.
      * @option template-repository Composer repository if package is hosted on a private registry or url to git.
-
+     * @option ci-template Git repo that contains the CI scripts that will be copied if there is no ci in the source project.
      */
     public function createProject(
-        // @todo: Add ci-template option.
         $source,
         $target = '',
         $options = [
@@ -263,6 +262,7 @@ class ProjectCreateCommand extends BuildToolsBase
             'visibility' => 'public',
             'region' => '',
             'template-repository' => '',
+            'ci-template' => '',
         ])
     {
         $this->warnAboutOldPhp();
@@ -276,6 +276,7 @@ class ProjectCreateCommand extends BuildToolsBase
         $visibility = $options['visibility'];
         $region = $options['region'];
         $use_ssh = $options['use-ssh'];
+        $ci_template = $options['ci-template'];
 
         // Provide default values for other optional variables.
         if (empty($label)) {
@@ -337,7 +338,7 @@ class ProjectCreateCommand extends BuildToolsBase
                 $version = $info->version();
                 $cms_version .= substr($version, 0, 1);
             }
-            $this->copyCiFiles($this->ci_provider, $siteDir, $cms_version);
+            $this->copyCiFiles($this->ci_provider, $siteDir, $cms_version, $ci_template);
 
             // If folder does not exists, assume we need to install composer deps.
             exec("composer --working-dir=$siteDir require --dev drupal/coder dealerdirect/phpcodesniffer-composer-installer squizlabs/php_codesniffer phpunit/phpunit");
--- a/src/Commands/ProjectCreateCommand.php
+++ b/src/Commands/ProjectCreateCommand.php
@@ -260,7 +260,7 @@ class ProjectCreateCommand extends BuildToolsBase
             'visibility' => 'public',
             'region' => '',
             'template-repository' => '',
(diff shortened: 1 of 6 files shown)


src/Commands/ProjectCreateCommand.php
```diff
@@ -139,6 +140,66 @@ public function validateCreateProject(CommandData $commandData)
...
+
+            case 'GithubActionsProvider':
+                return 'github';
+        }
+    }
+
+    /**
+     * Copy CI files from the given/default repo.
+     */
+    public function copyCiFiles($ci_provider, $created_folder, $cms_version = 'd8', $repo = '') {
```
kporras07, line 166: Pending: Add ci-repo option. @greg-1-anderson could you confirm that 'ci-repo' is a good name for the repo to use for the CI integrations?

src/Commands/ProjectCreateCommand.php
```diff
@@ -139,6 +140,66 @@ public function validateCreateProject(CommandData $commandData)
...
+        $short_name = $this->getCiShortName($ci_provider);
+        if (!$repo) {
+            $repo = 'git@github.com:kporras07/tbt-ci-integrations.git';
```
kporras07, line 169: I'll probably move this to default value unless some special logic is needed.

@greg-1-anderson should we put this repo now under pantheon-systems?

src/Commands/ProjectCreateCommand.php
```diff
@@ -426,6 +514,9 @@ function ($state) use ($site_name, $siteDir) {
             ->progressMessage('Install CMS on Pantheon site {site}', ['site' => $site_name])
             ->addCode(
                 function ($state) use ($ci_env, $site_name, $siteDir) {
+                    // Wait for workflow to finish.
+                    // @todo: Is this command working as expected?
+                    $this->passthru("terminus workflow:wait $site_name.dev 'Change database version for an environment'");
```
kporras07, line 519: @greg-1-anderson this command doesn't seem to be working as expected. Is there a better way to do this?

If I remove this, it randomly fails because sometimes the site installation is attempted before the db upgrade has been done. Also, waiting for N seconds doesn't seem to be reliable because it sometimes takes longer

src/Commands/ProjectCreateCommand.php
greg-1-anderson: How about `--ci-template`?

src/Commands/ProjectCreateCommand.php
greg-1-anderson: Yeah, do you have access rights to move it?

kporras07: I could transfer the repository and someone with enough access on pantheon-systems should accept it. Should I rename it before using this transfer option? I'm guessing you would be able to accept the transfer, is that right?

src/Commands/ProjectCreateCommand.php
```diff
@@ -139,6 +140,66 @@ public function validateCreateProject(CommandData $commandData)
         $this->providerManager()->validateCredentials();
     }
 
+    /**
+     * Get CI short name based on the CI provider.
+     */
+    public function getCiShortName($ci_provider) {
+        $reflect = new \ReflectionClass($ci_provider);
```
greg-1-anderson, line 147: Instead of reflection, maybe we should add a 'shortName' method to the provider, and call that?

src/Commands/ProjectCreateCommand.php
```diff
@@ -139,6 +140,66 @@ public function validateCreateProject(CommandData $commandData)
...
+                "./vendor/bin/phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md --ignore=node_modules,bower_components,vendor ./web/themes/custom",
+                "./vendor/bin/phpcs --standard=DrupalPractice --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md --ignore=node_modules,bower_components,vendor ./web/modules/custom",
+                "./vendor/bin/phpcs --standard=DrupalPractice --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md --ignore=node_modules,bower_components,vendor ./web/themes/custom",
+            ];
+            if ($cms_version === 'd8' || $cms_version === 'd9') {
+                $composer_json['extra']['build-env']['export-configuration'] = "drush config-export --yes";
+            }
+            file_put_contents("$created_folder/composer.json", json_encode($composer_json, JSON_PRETTY_PRINT));
+        }
+        $this->passthru("mkdir $created_folder/web/modules/custom");
```
greg-1-anderson, line 196: Prefer Symfony Filesystem methods to passthru(shell) wherever possible.

src/Commands/ProjectCreateCommand.php
```diff
@@ -355,6 +437,12 @@ function ($state) use ($site_name, $label, $team, $target, $siteDir, $region) {
             ->progressMessage('Make initial commit')
             ->addCode(
                 function ($state) use ($siteDir, $source) {
+                    if (file_exists("$siteDir/web/modules/custom/.gitkeep")) {
```
greg-1-anderson, line 440: We shouldn't force-add like this. The reason these directories are .gitignore'd is that there is a Composer project type "Drupal custom module" which places projects in web/modules/custom. Folks can change the .gitignore if they want to commit files here instead. Is there a reason you need this committed to the repo?

src/Commands/ProjectCreateCommand.php
greg-1-anderson: I think that `pushCodeToPantheon` is already doing a workflow:wait. Maybe the timing or workflow names changed in some way to make this less reliable.

src/Commands/ProjectCreateCommand.php
kporras07: I'll fix this by renaming the folders in the template repo. That's what makes most sense.

src/Commands/ProjectCreateCommand.php
kporras07: I need the folders to exist to run phpcs:
```
./vendor/bin/phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md --ignore=node_modules,bower_components,vendor ./web/modules/custom
```

With this, I'm just force-adding an empty file in order to have the custom folder, but every other file in that folder will still be ignored to follow the current pattern.

I could also make a PR to https://github.com/pantheon-systems/drupal-project/blob/default/.gitignore to exclude the .gitkeep file and avoid force adding like this.

Another option would be to change the phpcs command to run it only if the folder exists.

Please let me know what do you think it's the best option and I'll update the code accordingly.

src/Commands/ProjectCreateCommand.php
kporras07: Done. Thanks

kporras07:
@greg-1-anderson I fixed several items. There are only 2 open comments waiting for your input.  Please let me know how to proceed.

src/Commands/ProjectCreateCommand.php
kporras07: Done

kporras07:
@greg-1-anderson this is ready now. However, tests keep failing (same than main). Should I look into them?

greg-1-anderson (approved):

PR #396, opened 2021-09-16
CMS-254: Make Drupal 9 discoverable on build tools.

kporras07:
This PR allows using "d9" as source on project create.

I'll also create a docs PR for this.


greg-1-anderson (approved):
LGTM. Investigating test failures optional. We should also consider publishing the upstream to packagist, so that we can just run `composer create-project` on it, but we'll defer that decision.

kporras07:
I'll look at the tests in a separate PR. Merging now

PR #405, opened 2021-10-06
CMS-294: Cut lines should be at the top of the file.

kporras07:
In order to preserve consistency with the example repos, the cut mark should be at the top of the file because we need the full file for Pantheon and the cut behavior is "remove everything above this line".

So, gitignore additions only for the source repo should be added BEFORE the cut line as described in the instructions.


src/Commands/ProjectCreateCommand.php
```diff
@@ -468,8 +468,9 @@ function () use ($siteDir) {
                                 '# Put ignore patterns for css build artifacts and similar items below.',
```
greg-1-anderson, line 468: The comment needs to change too -- artifacts go above the line, so they are ignored in GitHub et. al., and committed on Pantheon.

greg-1-anderson (approved):

PR #411, opened 2021-11-10
ignore the platform requirements for behat-wordpress-extension

jazzsequence:
the wp behat extension requires php >7.2 <7.4. The only version that is compatible with our environment is 0.7.1 which just requires php >5.6.0
This extension is no longer supported but we definitely don't need an ancient version so we can ignore the php requirements to prevent other composer conflicts.

This one-line fix resolves an issue that was created when `roave/security-advisories` switched their default branch to `latest` from `master` and `roots/bedrock` updated to accomodate.


kporras07:
Looks good to me. This will go to 2.x version of this plugin. We should cherry-pick it to 3.x later.

kporras07 (approved):

jazzsequence:
I can create a new branch off 3.x and make a separate pr.

PR #412, opened 2021-11-10
ignore the platform requirements for behat-wordpress-extension

jazzsequence:
the wp behat extension requires php >7.2 <7.4. The only version that is compatible with our environment is 0.7.1 which just requires php >5.6.0
This extension is no longer supported but we definitely don't need an ancient version so we can ignore the php requirements to prevent other composer conflicts.

**cherry-picked from #411


kporras07 (approved):
:shipit:

PR #421, opened 2021-12-07
Make composer require commands run with no-update and then run composer update.

kporras07:
Also remove unneeded direct dependency (symfony/css-selector) already required by behat/mink.


greg-1-anderson (approved):
Looks good to get us back to running.

For consideration for possible future follow-on work: could we also get the dependencies to add from files copied via `copyCiFiles` above? Hardcoding isn't good, e.g. if we have different Drupal or WordPress project templates that need different dev dependencies.

Let's just merge like this for now, though.

PR #427, opened 2022-01-25
Fix regex for gitlab tokens.

kporras07:
This was previously 20 characters, now it's 26 characters. I'm opening support for both in case there are old gitlab installations still on 20 characters (not even sure if this is possible though)

### DIFF
--- a/.circleci/set-up-globals.sh
+++ b/.circleci/set-up-globals.sh
@@ -20,7 +20,8 @@ set -ex
 # Update terminus temporarily.
 rm /usr/local/bin/terminus
 mkdir ~/terminus && cd ~/terminus
-curl -L https://github.com/pantheon-systems/terminus/releases/download/`curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | perl -nle'print $& while m#"tag_name": "\K[^"]*#g'`/terminus.phar --output terminus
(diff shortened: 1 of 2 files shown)


.circleci/set-up-globals.sh
```diff
@@ -18,12 +18,11 @@ source $BASH_ENV
...
 
 # Update terminus temporarily.
-cd /opt/terminus
-git fetch
-git checkout 3.x
-composer install
 rm /usr/local/bin/terminus
-ln -s /opt/terminus/bin/t3 /usr/local/bin/terminus
+mkdir ~/terminus && cd ~/terminus
+curl -L https://github.com/pantheon-systems/terminus/releases/download/`curl --silent "https://api.github.com/repos/pantheon-systems/terminus/releases/latest" | perl -nle'print $& while m#"tag_name": "\K[^"]*#g'`/terminus.phar --output terminus
```
CdrMarks: Is it correct that there shouldn't be a space after `download/`?

CdrMarks: I guess so.  For readability, it seems that the nested curl statement could be on a line before this and then included in the download URL as a varible.

CdrMarks: I pushed this to address that: https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/427/commits/1cfa0b716c897393f22c2ad910df6501f5287a43

CdrMarks (approved):

greg-1-anderson (approved):

PR #429, opened 2022-01-28
Add githubactions to ci option description [ci skip].

kporras07:


CdrMarks (approved):

greg-1-anderson (changes requested):
Now it would make sense for github to default to githubactions, and this new help message implies that is the case. We probably shouldn't change the default until the next major release, though, so this message should probably explicitly mention that the default for GitHub is CircleCI.

greg-1-anderson (approved):

kporras07:
Addressed Greg's suggestion. Merging

PR #432, opened 2022-02-08
DB-1369: Added support to specify profile at the time of installation.

coderkakarrot:


lcatlett:
@coderkakarrot for this to work as intended, `$profile` needs to be added everywhere that $site_install_options` are used:
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/3.x/src/Commands/BuildToolsBase.php#L634
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/3.x/src/Commands/EnvInstallCommand.php#L27
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/3.x/src/Commands/ProjectCreateCommand.php#L565


Additionally, profile needs to be set as a site attribute so it can be available to the site environment provider and credential store (see account-name/admin-username for an example)	:
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/3.x/src/ServiceProviders/SiteProviders/PantheonProvider.php#L85
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/3.x/src/ServiceProviders/SiteProviders/PantheonProvider.php#L123
https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/3.x/src/ServiceProviders/SiteProviders/SiteEnvironment.php#L66-L75


Finally I don't think you want to set the default value as standard profile - this is optional so an empty string is fine because drupal will use standard if the profile option is not set by drush site-install

kporras07:
One more thing to keep in mind: what should happen if a Wordpress site receives that argument? Is there a similar concept in Wordpress? Should it be ignored or throw an error/exception?

backlineint:
> One more thing to keep in mind: what should happen if a Wordpress site receives that argument? Is there a similar concept in Wordpress? Should it be ignored or throw an error/exception?

I don't believe there is a similar concept in WordPress (I wish there was...) Is there an existing convention in build tools for anything that is CMS specific, or is this the first instance of that? 

Long term, Decoupled will actually need to develop an approach to choose between different mu-plugins depending on the front end framework being used. Given that, I wonder if not having build tools throw an exception if profile is provided could open up some options to solve that problem. In a way, we're really using mu-plugins as a stand in for install profiles. The other alternative would probably be adding an mu-plugin option of some kind.

backlineint:
I tested the latest version of this branch with the project create command we've been using to create a CMS backend using install profiles for decoupled. Confirmed that it can be used to install our `pantheon_decoupled_profile` and `pantheon_decoupled_umami_demo` profiles. Also confirmed that if --profile is not provided the site is installed using the standard profile.

backlineint (approved):

greg-1-anderson:
OK, seems fine. As follow-on work, it would be nice if an error or at least a warning were shown if `profile` was specified, but does not appear in the site install template. The existing code doesn't have this capability, so I am fine continuing without it for now.

Issue #434, opened 2022-03-04
Build Tools 3.x doesn't work with Terminus 3.0


aaronbauman:
Getting this error on `build:env:install`

```
Fatal error: Uncaught TypeError: preg_match(): Argument #2 ($subject) must be of type string, array given in /home/tester/.terminus/terminus-dependencies-617a5343ca/vendor/pantheon-systems/terminus-build-tools-plugin/src/Commands/BuildToolsBase.php on line 800
```

After terminus `auth:login`, I'm running this command within docker-build-tools-ci:8.x-php8.0
```
terminus build:env:install example.pr-123 --site-name="composer-base" --site-mail="test@example.com" --account-name=admin --account-mail="test@eample.com" --account-pass=password -v
```

Here's the full stacktrace
```
Call Stack:
    0.0069     459648   1. {main}() /usr/local/bin/terminus:0
    0.0393    1664600   2. require('phar:///usr/local/bin/terminus/bin/terminus') /usr/local/bin/terminus:12
    0.1258    7453528   3. Pantheon\Terminus\Terminus->run($input = ???, $output = ???) phar:///usr/local/bin/terminus/bin/terminus:78
[...]

PR #435, opened 2022-03-04
Input checking on call to escapeArgument; addresses #434

aaronbauman:
Addresses #434


kporras07:
Looks good to me. I confirm it works.

The issue was due to a (new?) option define (coming from T3) that could be provided multiple times and therefore is an array. With this, we're ignoring array options which I'd like to get @greg-1-anderson confirmation that it's ok before merging.

kporras07:
@aaronbauman thanks for the clear report and PR

aaronbauman:
Right, i should have stated that explicitly.
With this PR, the `--define` options don't get escaped. 

Wasn't sure if that was the best way to go or not, and couldn't figure it out from documentation.

PR #440, opened 2022-03-16
[CMS-637] Update infer code for gitlabci providers.

kporras07:
I'd like to get this PR approved but I'll hold on on merge until I can actually test it.

To test it I need gitlab to make the breaking change again.

See https://gitlab.com/gitlab-org/gitlab/-/issues/355946#note_877147666


greg-1-anderson:
I think that we could merge this. Might want to refactor into a single static method that both providers could use. Not sure where the best location for that is; it's probably okay to make a cross-provider call to a public static if there isn't a better util location for it.

kporras07:
I'll keep it open and will merge if the Gitlab issue comes back

PR #449, opened 2022-04-26
Allow config-based overrides of github server hostname

benjaminbradley:


src/ServiceProviders/RepositoryProviders/GitHub/GitHubProvider.php
```diff
@@ -25,9 +26,19 @@ class GitHubProvider extends BaseGitProvider implements GitProvider, LoggerAware
     protected $api;
 
 
+    public function __construct(Config $config)
+    {
+        parent::__construct($config);
+        $privateGithub = $this->config->get('build-tools.provider.git.github.url');
```
kporras07, line 32: Hi @benjaminbradley 

This looks great, but there's an underlying issue that makes this not work with env vars as described in this PR: https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/187

Could you please update this line to be:

```
        $privateGithub = $this->config->get('build-tools.provider.git.github.url') ?? getenv('TERMINUS_BUILD_TOOLS_PROVIDER_GIT_GITHUB_URL') ?? null;
```
so that this env var (TERMINUS_BUILD_TOOLS_PROVIDER_GIT_GITHUB_URL) can be used to override this behavior easily in CI? 

Thanks!

Once merged, I'll create a follow-up PR to set that var when creating a project from scratch and I will probably request help to test it, it would be great if you can help with that at that time :)

kporras07 (changes requested):

PR #453, opened 2022-06-17
Fix no-git-force option parameter

sergiuteaca:
The correct name of the parameter is `no-git-force` but it is sending `git-force`
The parameter name in the $options array in the method parameter list:
```
$options = [
            'label' => '',
            'clone-content' => false,
            'notify' => '',
            'db-only' => false,
            'message' => '',
            'pr-id' =>  '',
            'no-git-force' =>  false,
        ])
````
And lower is an example of how it is correclty used
```
$metadata = $this->pushCodeToPantheon($site_env_id, $multidev, '', $env_label, $options['message'], $options['no-git-force']);
```

### DIFF
--- a/src/Commands/EnvCreateCommand.php
+++ b/src/Commands/EnvCreateCommand.php
@@ -57,7 +57,7 @@ class EnvCreateCommand extends BuildToolsBase
         if ('dev' === $multidev) {
             $this->log()->notice('dev has been passed to the multidev option. Reverting to dev:env:push as dev is not a multidev environment.');
             // Run build:env:push.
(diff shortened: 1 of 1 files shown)


stevector:
@kporras07, can you review and merge? It looks like a straight-forward fix. And I'm able to replicate the problem. I think it hasn't been raised more because it show up only on builds for the `dev` environment.

kporras07 (approved):

namespacebrian (approved):

kporras07:
Merged, thanks!

PR #458, opened 2022-08-09
Fix deprecated trim() messages when using PHP 8.1

vinmassaro:
Running `build:project:create` using PHP 8.1, shows deprecated messages like:

` PHP Deprecated:  trim(): Passing null to parameter #1 ($string) of type string is deprecated in /Users/vincentmassaro/.terminus/terminus-dependencies-4f7bf5e1bb/vendor/pantheon-systems/terminus-build-tools-plugin/src/Credentials/CredentialManager.php on line 128`

This commit fixes these instances of `trim()` in `src/Credentials/CredentialManager.php` so it doesn't receive a null value. Thanks.

### DIFF
--- a/src/Credentials/CredentialManager.php
+++ b/src/Credentials/CredentialManager.php
@@ -125,7 +125,7 @@ class CredentialManager implements CredentialProviderInterface
             return;
         }
         $credential = $this->cache->get($key);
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Thanks!

PR #465, opened 2022-10-17
[CMS-1078] Tweak tests a bit.

kporras07:
1) Re-add image replacement script that is actually still needed.

### DIFF
--- a/.circleci/test-github-repair.sh
+++ b/.circleci/test-github-repair.sh
@@ -51,7 +51,7 @@ createTestPR 'test-after-repair' "Test after repair"
 # Create a bunch of pull requests that will not run any tests.
 # We do this so that we'll have to make more than one API request
 # to find our test PR.
-for n in $(seq 1 5) ; do
+for n in $(seq 1 10) ; do
     createTestPR "no-op-$n" "[ci skip] Pull request that is not tested (#$n)"
 done
 
--- a/.circleci/test-github-repair.sh
+++ b/.circleci/test-github-repair.sh
@@ -51,7 +51,7 @@ createTestPR 'test-after-repair' "Test after repair"
 # Create a bunch of pull requests that will not run any tests.
 # We do this so that we'll have to make more than one API request
 # to find our test PR.
-for n in $(seq 1 10) ; do
+for n in $(seq 1 5) ; do
     createTestPR "no-op-$n" "[ci skip] Pull request that is not tested (#$n)"
 done
 
--- a/.circleci/test-github-repo.sh
+++ b/.circleci/test-github-repo.sh
@@ -23,3 +23,10 @@ terminus build:project:create -n "$SOURCE_COMPOSER_PROJECT" "$TERMINUS_SITE" --t
 terminus site:info "$TERMINUS_SITE"
 # Confirm that the Github project was created
 git clone "$CLONE_URL" "$TARGET_REPO_WORKING_COPY"
+
+# Replace with right docker image.
+cd $TARGET_REPO_WORKING_COPY
+sed -i 's/quay.io\/pantheon\-public\/build\-tools\-ci:6.x/pantheonpublic\/build\-tools\-ci:8.x\-php7.4/g' ./.circleci/config.yml
+git add .circleci/config.yml
+git commit --allow-empty -m "Update Docker image if needed."
+git push origin master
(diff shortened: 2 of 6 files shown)


.circleci/test-github-repo.sh
```diff
@@ -23,3 +23,10 @@ terminus build:project:create -n "$SOURCE_COMPOSER_PROJECT" "$TERMINUS_SITE" --t
 terminus site:info "$TERMINUS_SITE"
 # Confirm that the Github project was created
 git clone "$CLONE_URL" "$TARGET_REPO_WORKING_COPY"
+
+# Replace with right docker image.
+cd $TARGET_REPO_WORKING_COPY
+sed -i 's/quay.io\/pantheon\-public\/build\-tools\-ci:6.x/pantheonpublic\/build\-tools\-ci:8.x\-php7.4/g' ./.circleci/config.yml
```
greg-1-anderson, line 30: Do we need to do this because there are other contexts where the fixture needs to use the 6.x version? If not, why not just update the fixture?

.circleci/test-github-repair.sh
```diff
@@ -51,7 +51,7 @@ createTestPR 'test-after-repair' "Test after repair"
 # Create a bunch of pull requests that will not run any tests.
 # We do this so that we'll have to make more than one API request
 # to find our test PR.
-for n in $(seq 1 10) ; do
+for n in $(seq 1 5) ; do
```
greg-1-anderson: Is five PRs enough to require a second API request to find the test PR? I thought that the API by default paged after ten items. If we're not forcing the second API call, then we might as well have zero base PRs.

greg-1-anderson: It looks like this happens in the test for build:project:repair; that whole section takes just under a minute. Maybe it isn't so costly to just make ten PRs?

.circleci/test-merge-master.sh
```diff
@@ -80,6 +80,7 @@ STATUS="$?"
 # Assert that status is non-zero
 if [ $STATUS -eq 0 ] ; then
     echo "Environment $TERMINUS_SITE.$TERMINUS_ENV should have been deleted, but was not."
-    exit 1
+    echo "This should fail the test but right now it is not working so marking as complete."
```
greg-1-anderson: Do you know why the delete environment is not working? I'd use the nomenclature of marking the test skipped if that's what we're going to do, but it would be better to fix if possible.

.circleci/test-github-repair.sh
kporras07: I was trying to reduce API requests to try to avoid hitting Github limit

.circleci/test-merge-master.sh
greg-1-anderson: Looks like the deletion worked in the last test run. Is this an intermittent issue? If so, maybe note that in the message describing why we're skipping the check.

kporras07: It is still creating the environment when we get here so the delete environment seems to be rejected at all. Problem is that the site uses IC so it takes way longer than before. These tests seems to have been broken/unstable for quite a long time.

I'd love to create a follow-up ticket for this that we (hopefully) tackle anytime soon(ish)

.circleci/test-github-repo.sh
kporras07: hmmm that's a good point. This is for WP-only (iirc); I guess I could just change it there if that makes sense

.circleci/test-github-repair.sh
kporras07: Just re-read and understood this. I'll go back to 10

.circleci/test-github-repo.sh
kporras07: @greg-1-anderson just double checking. This won't be in a fixture, it's the wordpress template project. I still think it should be ok to move it to 8.x-php7.4, just double checking. Please confirm

.circleci/test-merge-master.sh
kporras07: I think I fixed this

.circleci/test-github-repo.sh
kporras07: Doing this seems to actually break the behat_tests over there so, with that and having in mind that we have wordpress-composer-managed in the near(ish) future then I'd like to keep it here at least as of now. @greg-1-anderson any objections?

greg-1-anderson (approved):

PR #466, opened 2022-11-30
Add a create-before-push and git-commit-compare options for build:env:create

davereid-pfg:
Workflows that push code via CI tools do not always pass the commitChangesFile('HEAD', 'pantheon.yml') check because on branches, the HEAD commit may not contain the change to pantheon.yml, but a previous commit will on the same branch.


src/Commands/EnvCreateCommand.php
```diff
@@ -42,6 +44,8 @@ public function createBuildEnv(
             'message' => '',
             'pr-id' =>  '',
             'no-git-force' =>  false,
+            'create-before-push' => false,
+            'git-commit-compare' => 'HEAD',
```
kporras07: Hi @davereid-pfg looks great! Just one small request, could you please rename this option to `git-ref-compare`. I will go ahead and merge after that

Thanks!

davereid-pfg: Changed!

kporras07:
Merged. Thanks

PR #470, opened 2023-01-23
[BUGS-5651] Workaround issue with preUpdate script.

kporras07:


src/Commands/ProjectCreateCommand.php
```diff
@@ -380,8 +381,13 @@ public function createProject(
             // Install packages depending on the application.
             if ($app === 'Drupal') {
                 exec("composer --working-dir=$siteDir require --no-update --dev drupal/coder drupal/drupal-extension drupal/drupal-driver");
-                exec("composer --working-dir=$siteDir require --no-update drush-ops/behat-drush-endpoint");
+                exec("composer --working-dir=$siteDir require --no-update --dev drush-ops/behat-drush-endpoint");
                 exec("composer --working-dir=$siteDir require --no-update pantheon-systems/quicksilver-pushback");
+
+                if (Comparator::greaterThanOrEqualTo($version, '9.5.0')) {
```
greg-1-anderson: Would it be okay to just change this in the template projects? What if the platform.php version is already > 8.1.13? What about php_version in pantheon.yml?

kporras07: This is just a stopgap solution to allow it to run `composer update` without downgrading to 9.4.4 (which was happening because 9.5.0 raised PHP requirement to >=8 (or 8.1?)).

Once `composer update` runs, it will run the preUpdate code and will set the final version as expected.

>Would it be okay to just change this in the template projects?

Template project is drupal-composer-managed, not sure if we want to do that, my guess is not.

>What if the platform.php version is already > 8.1.13? What about php_version in pantheon.yml?

This made me think: why not running the already existing pre-update-script in drupal-composer-managed? That would be way better. In order to do that, I need https://github.com/pantheon-systems/drupal-composer-managed/pull/14 to be merged into default. I think that's probably the best solution.

kporras07: I found this to not be true unless we deploy to pantheon-upstreams which we don't want right now.

So I made a workaround here and it's working now

namespacebrian (approved):

PR #510, opened 2026-03-04
[BUGS-10991] Add WaitForCommit utility and use it for code sync

greg-1-anderson:
Claude's implementation of my idea, which is essentially the same as @stevector 's idea. The basic idea here is to make a new Terminus release with a utility class that the Build Tools plugin can use, and then make a new version of the Build Tools plugin that requires this new Terminus release. As an interim measure, though, I have copied the utility class here. This is untested.

## Summary

- Copies the `WaitForCommit` utility class from [pantheon-systems/terminus#2789](https://github.com/pantheon-systems/terminus/pull/2789) into this plugin under the `Pantheon\TerminusBuildTools\Utility` namespace.
- Replaces the workflow-description-based `waitForCodeSync` implementation with the new commit-SHA-based `WaitForCommit::waitForCommit()`, which matches workflows by the actual commit hash rather than by description string.
- Adds `RequestAwareInterface` / `RequestAwareTrait` to `BuildToolsBase` to provide the `Request` dependency needed by `WaitForCommit`.
- Gets the target commit SHA via `git rev-parse HEAD` from the current working directory.

## Test plan

- [ ] Verify `build:env:create` pushes code and correctly waits for the sync workflow to complete on an existing multidev
- [ ] Verify `build:env:merge` merges to dev and correctly waits for the sync workflow to complete
- [ ] Verify timeout behavior via `TERMINUS_BUILD_TOOLS_WORKFLOW_TIMEOUT` env var


jazzsequence:
@greg-1-anderson I'm wondering about the duplication. When the updated Terminus version is released, will the Build Tools plugin need to be updated to remove the duplicate and use the version in Terminus core?

jazzsequence:
Wraps `build:workflow:wait` around Terminus core `workflow:wait` functionality as well.
Testing revewaled that the proposed fix would solve the problem for `build:env:push` but `build:workflow:wait` would still use the old wait functionality. I've added code that just invokes `Pantheon\Terminus\Helpers\Utility\WaitForCommit` rather than maintaining a unique version of it. 

This release would need depend on the latest/next version of Terminus to work properly, but for workflows that are installing the latest of both, that shouldn't be an issue.

jazzsequence (approved):

namespacebrian (approved):

